PR #11, opened 2019-08-01
Fix typescript errors

xorye:
Fixes typescript errors, enabling the extension to be built with `vsce package`.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -36,11 +36,6 @@ const config = {
             exclude: /node_modules/,
             use: [{
                 loader: 'ts-loader',
-                options: {
-                    compilerOptions: {
-                        "module": "es6" // override `tsconfig.json` so that TypeScript emits native JavaScript modules.
-                    }
-                }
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -36,11 +36,6 @@ const config = {
             exclude: /node_modules/,
             use: [{
                 loader: 'ts-loader',
-                options: {
```
fbricon, line 39: not sure about that one. Why do you remove it?

xorye: According to the default comment, those options override the typescript config.

Although I am not entirely sure why, overriding the typescript config causes typescript to be unable to find modules like vscode and vscode-languageclient, despite the fact that they are installed. 

image

I know that the third error in that screenshot is masked by the current typescript config. I suspect that the current typescript config is what allows vscode and vscode-languageclient to be found? This error only happens with those modules only. 

Also, since vscode-xml didn't have those options, I just deleted them.

PR #18, opened 2019-08-06
Use endpoints provided by fabric8

xorye:
Project generation and extensions list is now being retrieved from endpoints provided by  [launcher-quarkus](https://github.com/fabric8-launcher/launcher-quarkus).

The format of the extensions given by the endpoints from launcher-quarkus were different than the format of the extensions from before. After getting the list of extensions from their endpoint, I decided to [convert](https://github.com/xorye/vscode-quarkus/blob/728e86030101a4a778f4367d66736347f1a196b0/src/utils/requestUtils.ts#L33) it to match the [QExtensions interface](https://github.com/xorye/vscode-quarkus/blob/728e86030101a4a778f4367d66736347f1a196b0/src/definitions/extension.ts#L20) we have currently.

### DIFF
--- a/src/addExtensions/addExtensions.ts
+++ b/src/addExtensions/addExtensions.ts
@@ -19,7 +19,7 @@ import { State } from "../definitions/state";
 import { ConfigManager } from "../definitions/configManager";
 import { MultiStepInput } from "../utils/multiStepUtils";
 import { pickExtensionsWithoutLastUsed } from "../generateProject/pickExtensions";
-import { QExtension } from "../definitions/qExtension";
+import { QExtension } from "../definitions/extension";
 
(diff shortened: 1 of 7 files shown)


xorye:
Rebased with master. 

Also, if your settings.json has a value for `quarkus.tools.starter.apiUrl`, set that value to `"https://code.quarkus.io/api/quarkus"`. If that value exists, then that value takes priority for the `apiUrl`, rather than [`DEFAULT_API_URL`](https://github.com/redhat-developer/vscode-quarkus/blob/82878e7f5190a7d1137ff22cd1b07b007641e806/src/definitions/projectGenerationConstants.ts#L18).

PR #28, opened 2019-08-12
Add ability to overwrite existing folder and choose another folder for project generation

### LINKED ISSUE #16: Create a prompt for the user to select a new destination folder if folder with the same name already exists
When creating a project using "Quarkus: Generate a Maven Project", if the user enters an artifact id as "demo", and chooses a destination folder which already contains a folder named "demo" the project generation wizard closes and the user is presented with this error:

image

At this point, the user would have to start the project generation wizard from the beginning, all over again. 

In a situation like this, rather than an error, Spring Initializr provides this prompt:  

image

This prompt allows the user to go back to the project generation wizard and choose another folder.

Having something like this for vscode-quarkus would be nice.

xorye:
Fixes #16 

The user can now overwrite the existing project, or choose another directory.

The new prompt looks like this:
image

### DIFF
--- a/src/generateProject/generationWizard.ts
+++ b/src/generateProject/generationWizard.ts
@@ -6,7 +6,7 @@
 import * as path from 'path';
 import * as fs from 'fs';
 
-import { window, Uri, commands } from 'vscode';
+import { window, Uri, commands, OpenDialogOptions } from 'vscode';
 
 import { ConfigManager } from '../definitions/configManager';
 import { MultiStepInput } from '../utils/multiStepUtils';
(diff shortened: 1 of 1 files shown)


src/generateProject/generationWizard.ts
```diff
@@ -151,6 +144,42 @@ export async function generateProject(configManager: ConfigManager) {
...
+    const overrideChoice: string = await window.showWarningMessage(MESSAGE_EXISTING_FOLDER, OPTION_OVERWRITE, OPTION_CHOOSE_NEW_DIR);
+    if (overrideChoice === OPTION_CHOOSE_NEW_DIR) {
+      directory = await showOpenDialogFolder({ openLabel: LABEL_CHOOSE_FOLDER });
+    } else if (overrideChoice === OPTION_OVERWRITE) {
+        break;
+    } else { // User closed the warning window
+      return;
+    }
+  }
+  return Uri.file(path.join(directory.fsPath, projectName));
```
fbricon, line 171: directory can be undefined

fbricon (changes requested):

src/generateProject/generationWizard.ts
```diff
@@ -151,6 +144,42 @@ export async function generateProject(configManager: ConfigManager) {
   });
 }
 
+async function getTargetDirectory(projectName: string) {
+  const MESSAGE_EXISTING_FOLDER = `Directory named '${projectName}' already exists in selected directory.`;
```
fbricon: `'${projectName}' already exists in selected directory.`

xorye:
@fbricon I've made changes. When the target directory is undefined (which I think, can only happen if the user clicks 'cancel' in the folder selection dialog), the wizard exits and displays this error message:

image

Note that 'canceled' is the preferred American spelling.

Is there a more proper way to handle an undefined target directory? Maybe a different error message?

src/generateProject/generationWizard.ts
```diff
@@ -151,6 +144,48 @@ export async function generateProject(configManager: ConfigManager) {
   });
 }
 
+async function getTargetDirectory(projectName: string) {
+  const MESSAGE_EXISTING_FOLDER = `${projectName}' already exists in selected directory.`;
```
fbricon: `'${projectName}' ... 1st single quote is missing

xorye:
Oops. Fixed.

fbricon:
Thanks!

PR #33, opened 2019-08-16
Start vscode-quarkus on Quarkus command or on opening application.properties

### LINKED ISSUE #32: vscode-quarkus eagerly starts when vscode opens
The vscode-quarkus extension starts as soon as vscode starts. This is **bad**. Should only be triggered by either one of the quarkus commands or when application.properties is opened.

xorye:
Fixes #32 

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
 	"activationEvents": [
 		"onCommand:quarkusTools.createMavenProject",
 		"onCommand:quarkusTools.addExtension",
-		"*"
+		"onLanguage:quarkus-properties"
 	],
 	"main": "./dist/extension",
 	"extensionDependencies": [
@@ -38,6 +38,18 @@
 		"javaExtensions": [
 			"./jars/com.redhat.quarkus.jdt.core.jar"
 		],
+		"languages": [
+			{
+				"id": "quarkus-properties",
+				"aliases": [
+					"Quarkus properties"
+				],
+				"filenames": [
+					"application.properties"
+				],
+				"configuration": "./properties-support/language-configuration.json"
+			}
+		],
 		"commands": [
 			{
 				"command": "quarkusTools.createMavenProject",
@@ -46,10 +58,6 @@
 			{
 				"command": "quarkusTools.addExtension",
 				"title": "Quarkus: Add extensions to current project"
-			},
-			{
-				"command": "quarkusTools.jdtls",
-				"title": "Quarkus: Invoke jdt.ls"
 			}
 		],
 		"configuration": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
 	"activationEvents": [
 		"onCommand:quarkusTools.createMavenProject",
 		"onCommand:quarkusTools.addExtension",
-		"*"
+		"onLanguage:properties"
```
fbricon: No that's too broad. Just `application.properties` files for now.

We might need to create a Quarkus application properties type like in https://github.com/spring-projects/sts4/blob/0714b6a69309789ac85441062698153b0ac4d9c6/vscode-extensions/vscode-spring-boot/package.json#L55-L66

fbricon (changes requested):

package.json
fbricon: quarkus application properties type might be explored in another PR. Will be interesting to see how it will conflicts with vscode-spring-boot, since they'll likely use the same triggers.

package.json
```diff
@@ -38,6 +38,12 @@
 		"javaExtensions": [
 			"./jars/com.redhat.quarkus.jdt.core.jar"
 		],
+		"languages": [{
+			"id": "quarkus-properties",
+			"extensions": [".properties"],
```
fbricon: nope, that's still too broad. If you open any .properties file, it's considered a quarkus properties file

fbricon:
You also need to apply a textmate grammar to the new file type (see https://github.com/spring-projects/sts4/blob/0714b6a69309789ac85441062698153b0ac4d9c6/vscode-extensions/vscode-spring-boot/package.json#L175-L177), as well as a language configuration (https://github.com/spring-projects/sts4/blob/0714b6a69309789ac85441062698153b0ac4d9c6/vscode-extensions/vscode-spring-boot/properties-support/language-configuration.json)

package.json
xorye: I think the spring boot extension does this as well.
image

I created an empty `application.properties` file in an empty folder and the spring boot extension loaded up:
image

fbricon: try with foo.properties

xorye: the spring boot extension doesn't start up with that name. VSCode only sees it as a Properties file. Is it possible to determine if we are in a Quarkus project without starting the extension and getting information from jdt.ls?

fbricon: It's complicated. We'd prolly need to issue a command to jdt.ls to determine if we're in a quarkus project, so we'd need to start vscode-quarkus. catch-22

xorye:
I've applied the TextMate grammar and the language configuration. The extension now starts whenever a quarkus command is run in the command palette, or if you open a file called `application.properties`. As a result, `foo.properties` does not trigger the extension.

Issue #40, opened 2019-08-20
Updating quarkus.tools.starter.apiUrl requires a restart


fbricon:
if you have a bas apiUrl, which makes generation fail, fixing it requires a vscode restart to be taken into account. This should not be necessary

PR #42, opened 2019-08-21
Don't save apiUrl if using default apiUrl

### LINKED ISSUE #23: Generating a project shouldn't save apiUrl in "quarkus.tools.starter"
The API endpoint is not something users should need to reuse. This is a system preference.
This caused #22

xorye:
Fixes #23

The `quarkus.tools.starter` configuration option currently looks like the following:
```
quarkus.tools.starter: {
    apiUrl: <STRING>
    defaults: {
        groupId: <STRING>
        artifactId: <STRING>
        ...
    }
}
```

How the defaults saving previously worked, was that when we wanted to save defaults, instead of only saving only the `quarkus.tools.starter.defaults` object, I had to save the whole `quarkus.tools.starter` object instead. I was unable to only update only the `defaults` part.

I've tried `workspace.getConfiguration('quarkus.tools.starter').update('defaults', <NEW DEFAULTS>, true);` and other variations of this, without success.

The line of code above runs with an error: `Error: Unable to write to User Settings because quarkus.tools.starter.defaults is not a registered configuration.`

This is because of how `quarkus.tools.starter` is defined in [`package.json`](https://github.com/xorye/vscode-quarkus/blob/323c7928dcd447a6be42bf31a587fd908ac0882f/package.json#L69).

This PR will save the apiUrl only if the apiUrl previously existed in `settings.json`, and doesn't save the apiUrl otherwise. This is done [here](https://github.com/xorye/vscode-quarkus/blob/323c7928dcd447a6be42bf31a587fd908ac0882f/src/definitions/configManager.ts#L57-L62).

It might be cleaner if we defined two configurations in `package.json` instead of one nested one: `quarkus.tools.starter`.
They could be `quarkus.tools.starter.api` and `quarkus.tools.starter.defaults`.
This separates saving for the apiUrl and defaults.
This would also improve user experience because defining `quarkus.tools.starter.api` separately allows the user to change apiUrl like the image below, without opening `settings.json`.
image

If having only one configuration is good enough for our uses, then this PR leaves that as is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,58 +66,54 @@
 					"description": "Traces the communication between VS Code and the Quarkus language server.",
 					"scope": "window"
 				},
-				"quarkus.tools.starter": {
+				"quarkus.tools.starter.api": {
+					"type": "string",
+					"default": "http://code.quarkus.io/api/quarkus",
+					"pattern": "https?://.+",
+					"description": "Quarkus API base URL",
+					"scope": "window"
+				},
+				"quarkus.tools.starter.defaults": {
 					"type": "object",
+					"default": {},
+					"description": "Default values for the project generation wizard.",
+					"scope": "window",
 					"properties": {
-						"apiUrl": {
+						"groupId": {
 							"type": "string",
-							"pattern": "https?://.+",
+							"pattern": "^[A-Za-z0-9_\\-.]+$",
 							"scope": "window",
-							"description": "Quarkus API base URL"
+							"description": "Default Group Id"
 						},
-						"defaults": {
-							"default": {},
-							"type": "object",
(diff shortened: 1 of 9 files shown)


fbricon:
+1 to moving the API url to quarkus.tools.starter.api. Make sure you set the default value in package.json:
"quarkus.tools.starter.api": "http://code.quarkus.io/api/quarkus"

xorye:
Excellent, I will make the changes

xorye:
I've updated it, but still, the way the settings.json file is being read in this PR is not good (it stores the relevant information from settings.json on startup). The PR I will soon make for #40 will fix this so that settings.json is accessed every time a value is needed.

PR #46, opened 2019-08-26
Consider cases where pom.xml is not located at the root of workspace folder

### LINKED ISSUE #36: Adding an extension assumes the pom.xml is at the root of the folder
When opening a folder `foo` containing a `bar` quarkus project, running the `add extension` command will execute `mvn quarkus:add-extension -Dextensions="some.extension"` in the `foo` directory instead of the `bar` one.
If the workspace contains multiple pom.xml files, then the command should propose the list of pom.xml to apply the change to, then execute the command (using the mvn wrapper if present) with the `-f path/to/selected/pom.xml` flag

xorye:
Fixes #36 

When testing this PR, please delete `vscode-quarkus/node_modules` and run `npm install`.
Microsoft copyright headers have been added in some files due to the fact that they originated from [vscode-maven](https://github.com/Microsoft/vscode-maven/releases). 

Additionally, please upgrade VS Code to version 1.37.0 or greater.

When adding extensions, the extension searches for all files named `pom.xml` under all opened workspace folders. If there is more than one file named `pom.xml`, there will be a quick pick menu
to select which pom.xml file the user wants to add Quarkus extensions to. 
image
EDIT: validation for the input box was not implemented. Will implement in this PR soon.

If a file named `pom.xml` does not exist under all opened workspace folders, this error message appears:
image

Next, the extension will attempt to find a maven wrapper file closest to the newly found `pom.xml` (check the folder containing `pom.xml`, if wrapper the file is not found, check parent directory, repeat.) The search will stop if it reaches the workspace folder, without having found the wrapper files.

If the wrapper files were found, they will be used to invoke the maven command necessary to add the Quarkus extensions. Currently, this only works under the assumption that the maven wrapper files and the corresponding `.mvn/` directory is located in the same location.
If the wrapper files are not found, `mvn` will be used.


fbricon:
can you please fix the conflicts?

xorye:
done
I will start fixing conflicts in the other PRs too

PR #48, opened 2019-08-27
Keep unix permissions when unzipping project zip

### LINKED ISSUE #44: Inconsistent file permissions in generated Quarkus Maven project.
When generating a Maven project by calling 'Quarkus: Generate a Maven project' in the command palette, the final project that is generated has different file permissions that the project files generated and extracted from the downloaded zip from https://code.quarkus.io/.

Permissions from project files generated from vscode-quarkus:
image

Permissions from project files downloaded from https://code.quarkus.io/ as a zip, then extracted:
image

Most notably, the project files generated from vscode-quarkus does not have execute permissions for the maven wrapper files. 

This is because when vscode-quarkus extracts the downloaded project zip, the file stream writer that is responsible for writing the extracted files onto the disk will revoke execute permissions by default. 

One possible solution would be to tell the writer to give certain permissions to all files being written:
```
const options = {
  path: <DESTINATION PATH>,
  getWriter: (pathObject) => {
    return Writer({ path: pathObject.path, mode: 0755});
  }
};

request(<ENDPOINT URL>).pipe(unzipper.Extract(options));
```
This code snippet results in these permissions:
image

This gives `755` permissions to all files being written, including files located in subdirectories, which might not be desired.
image

xorye:
Fixes #44

Permissions when unzipping manually:
image

Permissions when this PR unzips:
image

I've noticed permissions for non-directories are the same, but permissions for directories are slightly different. This PR gives write permissions to the group, while unzipping manually does not.

### DIFF
--- a/src/utils/requestUtils.ts
+++ b/src/utils/requestUtils.ts
@@ -14,18 +14,21 @@
  * limitations under the License.
  */
 
-import * as rp from 'request-promise';
-import * as request from 'request';
-import * as unzipper from 'unzipper';
+import * as request from 'request-promise';
+import * as yauzl from 'yauzl';
+import * as p from 'path';
+import * as fs from 'fs';
+
 import { QExtension, APIExtension } from '../definitions/extension';
 import { ProjectGenState } from '../definitions/projectGenerationState';
+import { Readable } from 'stream';
 
 export async function getQExtensions(apiUrl: string): Promise<QExtension[]> {
   const requestOptions = {
     uri: `${apiUrl}/extensions`,
     timeout: 30000
   };
-  return rp(requestOptions)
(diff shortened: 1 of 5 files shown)


src/utils/requestUtils.ts
```diff
@@ -14,18 +14,21 @@
  * limitations under the License.
  */
 
-import * as rp from 'request-promise';
-import * as request from 'request';
-import * as unzipper from 'unzipper';
+import * as request from 'request-promise';
+import * as yauzl from 'yauzl';
```
fbricon, line 18: yauzl should be explicitely added to the dependencies in package.json

xorye: updated

PR #50, opened 2019-08-28
Add documentation for "Extract properties from Quarkus deployment jars"

angelozerr:
See https://github.com/redhat-developer/quarkus-ls/pull/52

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
 information about the currently opened Quarkus project and returns
 the information to vscode-quarkus.  
 
+The information is computed by scanning all classes annotated with `@ConfigRoot` Quarkus annotation from all project classpath JARs and generating the proper Quarkus properties.
+
+For instance if the project has a `pom.xml` with this dependency:
+
+```xml
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-core-deployment</artifactId>
+</dependency>
+```
+
+The scanner will collect the [io.quarkus.deployment.ApplicationConfig](https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/ApplicationConfig.java) class annotated with `@ConfigRoot` and will generate a Quarkus property for each field:
+
+ * `quarkus.application.name` for the `name` field.
+ * `quarkus.application.version` for the `version` field.
+ 
+In addition, the Quarkus jdt.ls extension is able to manage `Quarkus deployment JAR`. 
+
+For instance if the project has a `pom.xml` with this dependency:
+
+```xml
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-hibernate-orm</artifactId>
+</dependency>
+```
+
+this JAR doesn't contain the classes annotated with `@ConfigRoot`. The JAR which contains those classes is `quarkus-hibernate-orm-deployment*.jar`. This information comes from 
+the `META-INF/quarkus-extension.properties` of the `quarkus-hibernate-orm*.jar` as a property:
+		 
+```
+deployment-artifact=io.quarkus\:quarkus-hibernate-orm-deployment\:0.21.1
+```
+
+Under the hood, the quarkus jdt.ls extension delegates the resolution (i.e. download) of those deployment jars to Maven, regardless of the user project's build system. So, a project depending on:
+
+```xml
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-hibernate-orm</artifactId>
+</dependency>
+```
+
+will generate some hibernate properties like:
+
+ * `quarkus.hibernate-orm.dialect`
+ * `quarkus.hibernate-orm.sql-load-script`
+ 
+despite the fact that the following dependency was not declared in the `pom.xml`:
+
+```xml
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-hibernate-orm-deployment</artifactId>
+</dependency>
+```
+ 
 Step 5. vscode-quarkus receives the project information and sends it
 to the Quarkus language server.  
 


xorye:
I am reading through this now

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
 information about the currently opened Quarkus project and returns
 the information to vscode-quarkus.  
 
+The information is computed by scanning all classes annotated with `@ConfigRoot` Quarkus annotation from the all JARs of the project classpath and generate the proper Quarkus properties.
```
xorye: ... with the `@ConfigRoot` Quarkus annotation from all project classpath JARs and generating the proper ...

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+
+For instance if the project have a `pom.xml` with this dependency:
```
xorye: For instance, if the project has a ...

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+
+```xml
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-core-deployment</artifactId>
+</dependency>
+```
+
+The scanner will collect the [io.quarkus.deployment.ApplicationConfig](https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/ApplicationConfig.java) class annotated with `@ConfigRoot` and it will generate for each fields a Quarkus property:
```
xorye: ... and will generate a Quarkus property for each field:

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+
+ * `quarkus.application.name` for `name` field.
+ * `quarkus.application.version` for `version` field.
```
xorye: ... for the `name` field.
... for the `version` field.

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+ 
+The Quarkus jdt.ls extension is able to manage `Quarkus deployment JAR`.
```
xorye: In addition, the Quarkus jdt.ls ...

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+```
+
+The scanner will collect the [io.quarkus.deployment.ApplicationConfig](https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/ApplicationConfig.java) class annotated with `@ConfigRoot` and it will generate for each fields a Quarkus property:
+
+ * `quarkus.application.name` for `name` field.
+ * `quarkus.application.version` for `version` field.
+ 
+The Quarkus jdt.ls extension is able to manage `Quarkus deployment JAR`. 
+
+For instance if the project have a `pom.xml` with this dependency:
```
xorye: For instance, if the project has a ...

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+
+```xml
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-hibernate-orm</artifactId>
+</dependency>
+```
+
+this JAR doesn't contain the classes annotated with `@ConfigRoot`. The JAR which contains those classes is `quarkus-hibernate-orm-deployment*.jar`. This information comes from 
+the `META-INF/quarkus-extension.properties` of the `quarkus-hibernate-orm*.jar` as property:
```
xorye: ... as a property:

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+		 
+```
+deployment-artifact=io.quarkus\:quarkus-hibernate-orm-deployment\:0.21.1
+```
+
+The download of this deployment JAR kind is done with maven. In other words, a project which declares
```
xorye: (first option) The download for this kind of deployment JAR is done with Maven.
or
(second option) The downloads for deployment JARs are done with Maven.

Does there exist different kinds of deployment jars that are not downloaded with Maven? If yes, the first option would be correct. If all deployment jars, in general, are downloaded with Maven, the second option is good.

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+    <artifactId>quarkus-hibernate-orm</artifactId>
+</dependency>
+```
+
+will generate some hibernate properties like:
+
+ * `quarkus.hibernate-orm.dialect`
+ * `quarkus.hibernate-orm.sql-load-script`
+ 
+without declaring in the pom.xml, this dependency:
```
xorye: I would suggest something like:

despite the fact that the following dependency was not declared in the `pom.xml`:

CONTRIBUTING.md
angelozerr: > Does there exist different kinds of deployment jars that are not downloaded with Maven? 

Graddle are their own download utilities, that's why I thought at first if it's a maven project I use maven utility to download JARs, if it's a graddle project I use graddle utility class to download JARs. After discussing with @fbricon we decide to use every time maven utility.

angelozerr:
@xorye is there something that you wish I fix?

xorye:
It would be good if @fbricon reads over it one last time

CONTRIBUTING.md
```diff
@@ -50,6 +50,63 @@ Step 4. The Quarkus jdt.ls extension receives the command, determines
...
+```
+
+this JAR doesn't contain the classes annotated with `@ConfigRoot`. The JAR which contains those classes is `quarkus-hibernate-orm-deployment*.jar`. This information comes from 
+the `META-INF/quarkus-extension.properties` of the `quarkus-hibernate-orm*.jar` as a property:
+		 
+```
+deployment-artifact=io.quarkus\:quarkus-hibernate-orm-deployment\:0.21.1
+```
+
+The downloads for deployment JARs are done with Maven. In other words, a project which declares:
```
fbricon: Under the hood, the quarkus jdt.ls extension delegates the resolution (i.e. download) of those deployment jars to Maven, regardless of the user project's build system. So, a project depending on:

PR #56, opened 2019-08-30
Store default extensions in a more compact way

xorye:
Default extensions are now saved in settings.json in a more compact way:
image

### DIFF
--- a/src/generateProject/pickExtensions.ts
+++ b/src/generateProject/pickExtensions.ts
@@ -13,12 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+import { QUARKUS_GROUP_ID } from '../definitions/wizardConstants';
+import { Config } from '../Config';
 import { MultiStepInput } from '../utils/multiStepUtils';
-import { QExtension } from '../definitions/extensionInterfaces';
+import { QExtension } from '../definitions/QExtension';
 import { State } from '../definitions/inputState';
 import { getQExtensions } from '../utils/requestUtils';
-import { Config } from '../Config';
 
 enum Type {
   Extension,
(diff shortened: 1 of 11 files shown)


fbricon:
please fix the conflict

src/generateProject/pickExtensions.ts
```diff
@@ -135,6 +128,16 @@ async function pickExtensions(
   }
 }
 
+function getDefaultExtensions(groupdIdArtifactId: string[], allExtensions: QExtension[]): QExtension[] {
```
fbricon: function getDefaultExtensions(defaultExtensionIds: string[], allExtensions: QExtension[]): QExtension[] {

src/utils/extensionsUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { QExtension, APIExtension } from '../definitions/extensionInterfaces';
+
+export function convertToQExtensions(extensions: APIExtension[]): QExtension[] {
+  return extensions.map((extension) => {
+    const semicolon: number = extension.id.indexOf(':');
```
fbricon: You need to ensure there's a colon. If not, then assume the default groupId is io.quarkus. Need to check for empty value.
Need to check the extensions are not the old version of extensions (i.e. before this PR)

fbricon:
@xorye sorry but there's a new conflict

src/utils/extensionsUtils.ts
xorye: Can we assume that the extension Ids from the endpoint, always contains a semicolon?

fbricon: Doesn't hurt to check

xorye:
I have made the changes.

> You need to ensure there's a colon. If not, then assume the default groupId is io.quarkus. Need to
> check for empty value.
> Need to check the extensions are not the old version of extensions (i.e. before this PR)

For the user input from settings.json, these checks are done [here](https://github.com/xorye/vscode-quarkus/blob/1a3a456fadadddca6b4902168c59d1a1f88a5bf2/src/generateProject/pickExtensions.ts#L133-L146).
For the extensions retrieved from the api endpoint, these checks are done [here](https://github.com/xorye/vscode-quarkus/blob/1a3a456fadadddca6b4902168c59d1a1f88a5bf2/src/definitions/QExtension.ts#L42-L56).

PR #60, opened 2019-09-04
Generate launch.json and tasks.json

### LINKED ISSUE #13: generate launch.json
Generate a 'launch.json' with at least one configuration to attach a debugger to quarkus started on dev mode.

xorye:
Fixes #13 

When the user creates a new Quarkus project with the "Quarkus: Generate a Maven project" command, the new Quarkus debug config will be created:
image

This allows the user to hit F5 to debug, and have the mvn quarkus:dev command run automatically.

When the user exits the debug session, they are presented with this modal:
image

Clicking 'Cancel' will close the modal and will not terminate the quarkus:dev task.
Clicking 'Yes' or 'No' will cause a new modal to appear:
image

For this modal, clicking 'Cancel' has the same effect as clicking 'No'.
If 'Yes' is clicked, the two modals will not show up again, and the quarkus:dev task will terminate or will not terminate (depending on what the user clicked earlier) everytime the debug session is closed.
If 'No' is clicked, the two modals will show up again when a debug session is closed.

The user can also set the behaviour in VS Code settings:
image

There is one problem I found, however. If the user follows these steps:
1. Open VS Code
2. Open Quarkus project folder, but don't open any files
3. Set the config to 'Always terminate'
image
4. Start the debug session with F5.
5. Close the debug session

In this case, we would expect that the quarkus:dev task to terminate, but it does not. This is because vscode-quarkus is not running. It is not running because the user did not invoke a vscode-quarkus command / did not open application.properties. To fix this, there needs to be a way to start the vscode-quarkus extension when debugging a Quarkus project.

### DIFF
--- a/src/debugging/terminateProcess.ts
+++ b/src/debugging/terminateProcess.ts
@@ -1,0 +1,107 @@
+/**
+ * Copyright 2019 Red Hat, Inc. and others.
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *     http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { QuarkusConfig, TerminateProcessConfig } from '../QuarkusConfig';
+import { DebugSession, TaskExecution, debug, window } from 'vscode';
+import { Disposable } from 'vscode-languageclient';
+import { getRunningQuarkusDevTasks } from '../utils/tasksUtils';
+
+export function createTerminateDebugListener(disposables: Disposable[]): Disposable {
+
+  return debug.onDidTerminateDebugSession(async (debugSession: DebugSession) => {
+
+    let quarkusDevTaskExe: TaskExecution;
+
+    try {
+      checkHasPreLaunchTask(debugSession);
+      quarkusDevTaskExe = await getPreLaunchTaskExecution(debugSession);
+    } catch (message) {
+      // debugger for a non Quarkus project has terminated
+      // don't display an error message
+      return;
+    }
+
+    const terminate: boolean = await getUserInputTerminate();
+
+    if (terminate) {
+      quarkusDevTaskExe.terminate();
+    }
+
+  }, null, disposables);
+}
+
+function checkHasPreLaunchTask(debugSession: DebugSession): void {
+  if (typeof debugSession.configuration.preLaunchTask === 'undefined') {
+    throw 'Pre launch task does not exist';
+  }
+}
+
+async function getPreLaunchTaskExecution(debugSession: DebugSession): Promise<TaskExecution> {
+
+  const preLaunchTask: string = debugSession.configuration.preLaunchTask;
+
+  // Get quarkus:dev tasks from the debugSession's workspace folder
+  const runningQuarkusDevTasks: TaskExecution[] = await getRunningQuarkusDevTasks(debugSession.workspaceFolder);
+
+  for (const taskExe of runningQuarkusDevTasks) {
+    if (taskExe.task.name === preLaunchTask) {
+      return taskExe;
+    }
+  }
+  throw `Task named ${preLaunchTask} is not a running quarkus:dev task in closed debug session's WorkspaceFolder`;
+}
+
+function saveToConfig(terminate: boolean, askAgain: boolean): void {
+  let save: TerminateProcessConfig;
+  if (askAgain) {
+    save = TerminateProcessConfig.Ask;
+  } else if (terminate) {
+    save = TerminateProcessConfig.Terminate;
+  } else if (!terminate) {
+    save = TerminateProcessConfig.DontTerminate;
+  }
+  QuarkusConfig.saveTerminateProcessOnDebugExit(save);
+}
+
+async function getUserInputTerminate(): Promise<boolean> {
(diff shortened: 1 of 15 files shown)


xorye:
> In this case, we would expect that the quarkus:dev task to terminate, but it does not. This is because vscode-quarkus is not running. It is not running because the user did not invoke a vscode-quarkus command / did not open application.properties. To fix this, there needs to be a way to start the vscode-quarkus extension when debugging a Quarkus project.

There is an activation event for debugging: `onDebugResolve:type`, where `type` is the debug configuration type (ie, java, python, etc). Of course, if we have `onDebugResolve:java`, vscode-quarkus will start when debugging any java application, which is not what we want.

src/generateProject/createDebugConfig.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (insertSpaces) {
+    const numSpaces: number = workspace.getConfiguration('editor').get('tabSize');
+    tasksContent = tasksContent.replace(/\t/g, ' '.repeat(numSpaces));
+    launchContent = launchContent.replace(/\t/g, ' '.repeat(numSpaces));
+  }
+  fs.writeFileSync(vscodeDir + '/tasks.json', tasksContent);
+  fs.writeFileSync(vscodeDir + '/launch.json', launchContent);
+}
+
+const tasksJson: string =
```
xorye: There may be a better way to set tasks.json and launch.json through the VS Code Api. Currently investigating.

fbricon:
If I disconnect the debugger, I get no popup

fbricon:
Ok so it didn't start the vscode-quarkus extension, so no popup was invoked

fbricon:
For the activation issue, I'm afraid we're gonna need to add a new Quarkus command, that'll basically look for and execute the launch configuration, it's the only way to guarantee the extension will register its debug listener. I've tried to add "workspaceContains:**/target/wiring-devmode" as an activation event, but it only works if the workspace already contains that directory. Activating on all java debug events is excessive.

src/generateProject/createDebugConfig.ts
fbricon: I'd read embedded files, it's wayyy easier to maintain

xorye:
I've added a new command called "Quarkus: Debug current Quarkus project" that starts the debugger, but for now, it assumes that the project is located in the [first open workspace folder](https://github.com/xorye/vscode-quarkus/blob/ac0c90d7d94eccb940c4b3887277cca435c3e109/src/extension.ts#L89).

If the user has multiple folders open with multiple Quarkus projects open, we need to determine which project to debug. 

I've also updated the PR so that the tasks, launch json content are read from files.

src/debugging/terminateProcess.ts
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Config, TerminateProcessConfig } from '../Config';
+import { Disposable } from 'vscode-languageclient';
+import { TaskExecution, debug, tasks, window, workspace } from 'vscode';
+
+export function createTerminateDebugListener(disposables: Disposable[]): Disposable {
+  return debug.onDidTerminateDebugSession(async () => {
```
fbricon: check the debug session event is a quarkus debugging session. Else return early

xorye:
I've made the updates.

Changed the modal to an Information message and changed the button names: 
image

image

The change from modal to information message was because the button layout for modals seem to be platform [dependent](https://github.com/microsoft/vscode/issues/71251#issuecomment-477297944). I could not control where the 'Cancel' button was being placed.

Maybe [vscode #47899](https://github.com/Microsoft/vscode/issues/47899) would help us create input boxes with checkboxes in the future.

A problem with this PR is that we assume that the user would never change the `quarkus:dev` task label in tasks.json.

fbricon:
I think it looks better. However, since we now know the user's choice, we could be more specific on the 2nd popup:
Do you want to always [terminate the task|keep the task running]? Remember my choice / Ask again next time

xorye:
Made the changes, the second information message looks like this:
image

xorye:
This is the `DebugSession` object given from the `onDidTerminateDebugSession()` function:
image

Since the user is able to change `preLaunchTask` or change `name`, my idea was to iterate through all tasks to find tasks that call the maven `quarkus:dev` command. I can retrieve all tasks through the VS Code api with `tasks.fetchTasks()`. Once I find those tasks, I can find the correct task names and compare them with `preLaunchTask`. @fbricon what do you think?

fbricon:
I can't think of a more surefire way to identify that task so yeah you can try

xorye:
I've made the changes. Since there were so many cases, I opted for a [try catch](https://github.com/redhat-developer/vscode-quarkus/blob/a9f5c4b0896623b6e232beb8205dcf72bdbedf10/src/debugging/terminateProcess.ts#L26-L31) return` checks. 

First, I check if there if the debug session has a prelaunch task.
Then I gather all quarkus:dev tasks that are running AND defined in the current debugSession's workspace folder, into a list.
Then, I just compare preLaunchTask with the contents of the list.

fbricon:
@xorye can you fix the conflict and merge?

xorye:
updated

xorye:
sorry, please do not merge this yet. I must fix a problem that arises when the extension runs after being installed from a .vsix.

xorye:
Fixed. The problem was that the files that we being read for tasks.json and launch.json were located in the vscode-quarkus/src directory. I moved them out to a new vscode-quarkus/vscode-debug-files directory.

fbricon:
So if you try to debug a gradle-based project (any simple gradle java project), nothing happens. there should be a popup indicating why command could not run

xorye:
New error messages have been implemented.
image
(This message above is what appears when a gradle project is open and the debug command is called)


image


image

xorye:
There are some TODO comments that I've added. I can make the github issues to address them after this is merged.

fbricon:
Nice work @xorye !

PR #67, opened 2019-09-06
Add forward slash before path for snippets

xorye:
Adds a forward slash for the path.
image


fbricon (approved):
It's not necessary(see https://github.com/quarkusio/quarkus-quickstarts/blob/2cd8f7d2174b177d2fe663c75c30a82b7862e1f7/hibernate-orm-resteasy/src/main/java/org/acme/hibernate/orm/FruitResource.java#L22) but it's ok

PR #77, opened 2019-09-12
Update project generation input validation

xorye:
This PR updates the input validation for when you go through the project generation wizard.

image

groupId: same as https://code.quarkus.io/
artifactId: same as https://code.quarkus.io/
project version: alphanumeric characters and these symbols: ._-
package name: same as https://code.quarkus.io/
resource name: must start with a letter (capital or lowercase), then, alphanumeric characters and these symbols: ._-

### DIFF
--- a/src/generateProject/validateInput.ts
+++ b/src/generateProject/validateInput.ts
@@ -1,0 +1,61 @@
+/**
+ * Copyright 2019 Red Hat, Inc. and others.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// For validation functions, return error message if there is an error
+// return undefined otherwise
+
+export async function validateGroupId(userInput: string): Promise<string|undefined> {
+  // regex referenced from
+  // https://github.com/quarkusio/code.quarkus.io/blob/master/src/main/frontend/src/code-quarkus/pickers/info-picker.tsx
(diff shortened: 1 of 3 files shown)


xorye:
The regexes from the validation from https://code.quarkus.io/, are specified in this file: https://github.com/quarkusio/code.quarkus.io/blob/master/src/main/frontend/src/code-quarkus/pickers/__tests__/__snapshots__/info-picker.spec.tsx.snap

fbricon:
That's good, but how do we handle server-side validation errors? We might lacking behind the server at some point, miss some validation. How are those server side errors managed?

xorye:
If there is a server side error or a server side validation error, the request() fails and I am able to catch it. I'll implement a warning message

xorye:
The following message appears when the server doesn't give us a project, which could happen due to many reasons (if we give it invalid parameters, if server is down, if we cannot connect to wifi etc.):

image

xorye:
This PR is still not ready yet. I must continue testing. I'll post an update when things are good.

xorye:
I think its good now. 

This PR now consists of two commits:
The first commit implements the error message that appears when a project cannot be downloaded from the endpoint for any reason.

The second commit implements client side validation for the inputs from the project generation wizard. It has been updated to match the new changes specified here: https://github.com/quarkusio/code.quarkus.io/pull/142#event-2632554602.

src/generateProject/validateInput.ts
```diff
@@ -0,0 +1,61 @@
...
+  // https://github.com/quarkusio/code.quarkus.io/blob/master/src/main/frontend/src/code-quarkus/pickers/info-picker.tsx
+  const re = new RegExp('^([a-zA-Z_$][a-zA-Z\\d_$]*\\.)*[a-zA-Z_$][a-zA-Z\\d_$]*$');
+  if (!re.test(userInput)) {
+    return 'Invalid package name';
+  }
+  return undefined;
+}
+
+export async function validateResourceName(userInput: string): Promise<string|undefined> {
+  const re = new RegExp('^[A-Za-z]+[A-Za-z0-9_\\-.]*$');
```
fbricon: `-` is not allowed in class names

fbricon (changes requested):

PR #78, opened 2019-09-12
Update defaults for project generation

### LINKED ISSUE #75: defaults for quarkus new maven projects are not "proper"
- [x] (minor) Group id is "org.my.group" - suggestion: "org.acme" as used in quarkus getting started guide.
- [x] Instead of mixed case for project "MyQuarkusProject" use what is in quarkus default guides "getting-started" or "quarkus-getting-started". 
- [x] package name "PackageName" is mixed case where packages are normally lower case and prefixed. suggestion: use "org.acme" as in quarkus getting started guide or even better default to what user used for groupid.
- [x] Resource name would be better if could be "GreetingResource" as in Getting started guide.

xorye:
Fixes #75 


fbricon (changes requested):
I would make it more dynamic, i.e if user sets custom groupid, then default the package name to the same

xorye:
The change has been made.

PR #82, opened 2019-09-17
Generate launch.json tasks.json when needed

### LINKED ISSUE #80: Only generate tasks/launch config when needed
The tasks and launch configuration should be generated only when needed, i.e. when executing the Quarkus debug command. This would allow for quarkus projects not generated by the vscode-quarkus wizard to run as well.

xorye:
Fixes #80 

### What this PR does:
When the debug command is called and the debug configuration that starts the quarkus:dev task could not be found, this PR will try to add a new debug configuration (to launch.json) and new quarkus:dev task (to tasks.json).

If the launch.json file already exists, the new debug configuration will be appended to the `configurations` array in the existing launch.json file. If the launch.json file does not exist, a new launch.json file containing the new debug configuration will be created.

Same goes for the quarkus:dev task and the tasks.json file. If the tasks.json file exists, it will append the new task into the `tasks` array. If not, the file is created.

The reason why I added the comment-json npm module was because I wanted to parse a JSON string that contained comments within the JSON. In other words, I wanted to read [these](https://github.com/xorye/vscode-quarkus/tree/80generate_launch_needed/vscode-debug-files) files as a string, then parse it into a javascript object.

### DIFF
--- a/src/debugging/createDebugConfig.ts
+++ b/src/debugging/createDebugConfig.ts
@@ -16,34 +16,59 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
-import { Uri, workspace } from 'vscode';
+import { DebugConfiguration, TaskDefinition, Uri, workspace, tasks } from 'vscode';
+import { parse } from 'comment-json';
 
-export function createDebugConfig(directory: Uri) {
+export async function createDebugConfig(directory: Uri) {
 
   const vscodeDir: string = directory.fsPath + '/.vscode';
   if (!fs.existsSync(vscodeDir)) {
     fs.mkdirSync(vscodeDir);
   }
 
-  let tasksContent: string = getDebugConfigFile('tasks.json');
-  let launchContent: string = getDebugConfigFile('launch.json');
-
-  const insertSpaces: boolean = workspace.getConfiguration('editor').get('insertSpaces');
-  if (insertSpaces) {
-    const numSpaces: number = workspace.getConfiguration('editor').get('tabSize');
-    tasksContent = replaceTabsWithSpaces(tasksContent, numSpaces);
-    launchContent = replaceTabsWithSpaces(launchContent, numSpaces);
+  if (fs.existsSync(vscodeDir + '/tasks.json')) {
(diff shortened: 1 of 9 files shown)


xorye:
Not ready yet.

xorye:
Should be good now.

I had a problem where [`await tasks.fetchTasks()`](https://github.com/microsoft/vscode/blob/86baf839044cd6a886716eb0924bee6d23592f50/src/vs/vscode.d.ts#L5497) from the VSCode api wasn't getting the tasks when the tasks.json was recently created using `fs.writeFileSync()`. 

Now, instead of using `await tasks.fetchTasks()` to get all task names, I am using `fs.readFileSync()` to directly read the tasks.json file.

I had a similar problem with launch.json, which I fixed in the same way.

fbricon:
tslint fails https://travis-ci.org/redhat-developer/vscode-quarkus/jobs/586252796#L555

xorye:
Fixed

src/debugging/createDebugConfig.ts
```diff
@@ -16,34 +16,60 @@
...
+  } else {
+    let launchContent: string = getDebugResourceAsString('launch.json');
+    launchContent = replaceWithSpacesIfNeeded(launchContent);
+    fs.writeFileSync(vscodeDir + '/launch.json', launchContent);
   }
+}
 
-  fs.writeFileSync(vscodeDir + '/tasks.json', tasksContent);
-  fs.writeFileSync(vscodeDir + '/launch.json', launchContent);
+function getDebugResource<T>(filename: string): T {
```
fbricon, line 54: or `return parse(getDebugResourceAsString(filename));`

fbricon (changes requested):

src/debugging/createDebugConfig.ts
xorye: done

fbricon:
I was able to successfully launch some projects from https://github.com/quarkusio/quarkus-quickstarts. Nice job @xorye!

PR #89, opened 2019-09-20
Updated readme

maxandersen:
As discussed with @bob-davis


Bob-Davis:
> As discussed with @Bob-Davis

Ack - Publisher will be Red Hat, "for VS Code" will be in the name.

PR #97, opened 2019-09-27
Debug startup command uses mvn if maven wrapper unavailable

### LINKED ISSUE #95: Debug startup command to use mvn if mvnw does not exist
Hi everyone!

In a project where maven wrapper artifacts where removed for just cleanup/simplicity purposes, trying to use `Quarkus: Debug current Quarkus project` end up with (no surprize, of course):

```bash
> Executing task: ./mvnw quarkus:dev <

/bin/bash: ./mvnw: No such file or directory
The terminal process terminated with exit code: 127

Terminal will be reused by tasks, press any key to close it.
```

I would see it useful to consider using `mvn` instead of `mvnw` for such cases.

I know that otherwise I could create my own (custom) task at the project level for such need.

xorye:
Fixes #95

Fixing #95 required some refactoring, most notably the code now uses [WorkspaceConfiguration.update()](https://github.com/microsoft/vscode/blob/6f070c5a354d85afdcd3657b348d8c4d6a1269d4/src/vs/vscode.d.ts#L4183) to create launch.json or tasks.json if those files don't exist.
An example of this is here:

https://github.com/xorye/vscode-quarkus/blob/5331a7dad73bc1c857bb138bbe5b03263b2d008c/src/debugging/createDebugConfig.ts#L91-L92

The benefit of this is that the workspace configuration is being updated right away, which means that we can access tasks.json and launch.json using VSCode API lile so:

```
// get launch config
workspace.getConfiguration('launch', vscode.window.activeTextEditor.document.uri); 

// get tasks
tasks.fetchTasks();
```

Before this PR, launch.json and tasks.json were being created using `fs.writeFileSync()`, which made it so that the workspace configuration wasn't updated right away. Therefore the new tasks/cofigurations weren't immediately accessible with the VSCode API. Instead, the previous code read launch.json and tasks.json manually with `fs.readFileSync()`.

This PR isn't tested on Windows yet.


xorye:
I will be testing on Windows soon

xorye:
This PR works on windows.
This PR also considers cases where mvnw exists and mvnw.cmd doesnt (and vice versa).

src/debugging/CreateDebugConfig.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    DebugConfigCreator.debugConfigCreator.createVSCodeDirIfMissing();
+    await DebugConfigCreator.debugConfigCreator.addDebugTask();
+    await DebugConfigCreator.debugConfigCreator.addDebugConfig();
+  }
+
+  private constructor(directory: Uri) {
+    this.workspaceDir = directory;
+    this.dotVSCodeDir = directory.fsPath + '/.vscode';
+    this.tasksJsonDir = this.dotVSCodeDir + '/tasks.json';
```
fbricon: not a directory. so tasksJsonFile

src/debugging/CreateDebugConfig.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.launchJsonDir = this.dotVSCodeDir + '/launch.json';
```
fbricon: not a directory. so launchJsonFile

fbricon (changes requested):

xorye:
Fixed

PR #98, opened 2019-09-30
Definition from application.properties to Java field which have Config* annotation

angelozerr:
Definition from application.properties to Java field which have Config* annotation

See https://github.com/redhat-developer/quarkus-ls/issues/4

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,7 +15,7 @@
  */
 import * as requirements from './languageServer/requirements';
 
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -24,6 +24,7 @@ import { createTerminateDebugListener } from './debugging/terminateProcess';
 import { generateProjectWizard } from './generateProject/generationWizard';
 import { prepareExecutable } from './languageServer/javaServerStarter';
 import { tryStartDebugging } from './debugging/startDebugging';
+import { stringify } from 'querystring';
```
fbricon: unused

angelozerr: @fbricon removed

Issue #53, opened 2019-08-29
Add welcome page with Quarkus information


gorkem:
Can we add a welcome page with links to quarkus tutorials and information?

angelozerr:
Github settings provide the capability to set a website (next to description), I think we should set the link https://quarkus.io/ 

I cannot do that since I'm not administrator of this github project. Here a sample of link:

![image](https://user-images.githubusercontent.com/1932211/63978799-95a5ae80-cab7-11e9-9bf0-0faa80873640.png)

fbricon:
I believe @gorkem means a welcome page in the quarkus extension, similar to the one for the java extension pack

angelozerr:
Fixed with https://github.com/redhat-developer/vscode-quarkus/pull/91

PR #99, opened 2019-09-30
Replace vscode dependency with  @types/vscode

xorye:
@types/vscode provides a more up-to-date VS Code API that is needed for  #53 

The postinstall script was removed from package.json, because it is unecessary for @types/vscode. Source: https://github.com/microsoft/vscode/issues/70175, https://github.com/microsoft/vscode-iot-workbench/issues/596


xorye:
Fixing the cause for failed travis test now

xorye:
Fixed

PR #106, opened 2019-10-10
Add spacing setting for formatting

xorye:
Added formatting spacing setting for https://github.com/redhat-developer/quarkus-ls/pull/110

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,10 +167,16 @@
             }
           }
         },
+        "quarkus.tools.formatting.surroundEqualsWithSpaces": {
+          "type": "boolean",
+          "default": "false",
+          "description": "Insert spaces around the equals sign when formatting the application.properties file",
+          "scope": "window"
+        },
         "quarkus.tools.symbols.showAsTree": {
           "type": "boolean",
           "default": "true",
-          "description": "Show Quarkus properties as tree",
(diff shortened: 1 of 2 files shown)


xorye:
I've set the default to false (as in, don't add spaces around equals sign), since we didn't have spaces around the equals sign for completion.

angelozerr:
@xorye  could you resolve the conflicts please.

package.json
```diff
@@ -157,10 +157,16 @@
             }
           }
         },
+        "quarkus.tools.formatting.insertSpaces": {
```
angelozerr: Should not be `quarkus.tools.formatting.insertSpacesBetweenAssign` ?

fbricon: *.surroundEqualsWithSpaces ?

xorye: fixed

angelozerr: @xorye it seems @fbricon would prefer `quarkus.tools.formatting.surroundEqualsWithSpaces`? If it that please do the fix and do the same thing for Quarkus LS settings.

xorye: fixed with @fbricon 's suggestion

xorye:
The setting has now been updated to: `quarkus.tools.formatting.surroundEqualsWithSpaces`

PR #109, opened 2019-10-15
Add tests that check if extension activates when a `quarkus.tools` command is called

### LINKED ISSUE #64: Tests for ensuring extension activates
- [x] Make sure extension activates when calling Quarkus commands (Quarkus: Generate a Maven project, etc.)
- [ ] Make sure extension activates when application.properties file opens

xorye:
Fixes #64 (without the application.properties activation test)

This PR has two commits because, In addition to adding tests, I have a commit that removes [path-exists](https://www.npmjs.com/package/path-exists) as a dependency.

Previously, path-exists was used for the `path-exists.sync()` function. This PR removes path-exists and uses `fs.existsSync()`, since it has been undeprecated in Node.js since 6.8.0, vscode-quarkus uses version 10.11.0.

I also had an additional test to check that the quarkus commands would correctly register once the extension activates. This test is not in this PR, but maybe it can be added in the future:
```
 // manually activate vscode-quarkus
 it('should activate', () => {
   return vscode.extensions.getExtension('redhat.vscode-quarkus').activate();
 }).timeout(1 * 60 * 1000);

 // check if all quarkus commands are registered
 it('should register all Quarkus commands', () => {
   return vscode.commands.getCommands(true).then((commands) => {
     const foundQuarkusCommands = commands.filter((value) => {
       return QUARKUS_COMMANDS.indexOf(value) >= 0 || value.startsWith('quarkusTools.');
     });
     expect(foundQuarkusCommands.length).to.equal(QUARKUS_COMMANDS.length, 'Some Quarkus commands are not registered properly or a new command is not added to the test');
   });
 });
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -292,7 +292,6 @@
     "fs-extra": "^8.0.1",
     "glob": "^7.1.4",
     "md5": "^2.2.1",
-    "path-exists": "^4.0.0",
     "request": "^2.88.0",
     "request-promise": "^4.2.4",
     "user-home": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -261,6 +261,7 @@
     "build-ext": "./node_modules/.bin/gulp buildExtension"
   },
   "devDependencies": {
+    "@types/chai": "^4.2.3",
     "@types/fs-extra": "^7.0.0",
     "@types/md5": "^2.1.33",
     "@types/mocha": "^5.2.6",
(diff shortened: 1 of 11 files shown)


xorye:
In order to run the tests, run `npm run test`

package.json
```diff
@@ -261,13 +261,15 @@
     "build-ext": "./node_modules/.bin/gulp buildExtension"
   },
   "devDependencies": {
```
fbricon, line 263: npm detected a vulnerability. Please run `npm audit fix` as part of this PR

xorye: I pushed changes

fbricon:
I don't see the previous extensions anymore, always 0

xorye:
Are you talking about the last used extensions when generating a project?

fbricon:
opened #138

PR #111, opened 2019-10-15
Validate initial input value, offers context on what is wrong

### LINKED ISSUE #84: Validate initial input value, offer context on invalid values
When I run generate maven project the default value for artifact id is "MyQuarkusProject" and that gives two problems:

1) when pressing enter on it the wizard does not move forward nor does it highlight the artifact id is not allowed.
2) only when editing the field do you get an error highlighting the issue, but it just says "invalid artifact id" giving no hint on what is wrong - in this case seem to be that it has uppercase letters. 

So would be good to have the validation error shown immediately or at least when pressing Enter and then tell user what is actually wrong with the artifact id or say what a valid one would be.

xorye:
Fixes #84

The create project wizard will now validate the default value injected in the field:
validation gif

The error messages for invalid groupId, artifactId, package name etc. now provide context:
image
image

Also, I've noticed that there doesn't seem to be version validation for https://code.quarkus.io/. Is version something that should be validated for Maven projects in general?

Before, the version was validated against this regex:  `[A-Za-z0-9_-.]+$`

### DIFF
--- a/src/generateProject/validateInput.ts
+++ b/src/generateProject/validateInput.ts
@@ -14,48 +14,60 @@
 // For validation functions, return error message if there is an error
 // return undefined otherwise
 
-export async function validateGroupId(userInput: string): Promise<string|undefined> {
+export async function validateGroupId(userInput: string): Promise<string | undefined> {
   // regex referenced from
   // https://github.com/quarkusio/code.quarkus.io/blob/master/src/main/frontend/src/code-quarkus/pickers/info-picker.tsx
   const re = new RegExp('^([a-zA-Z_$][a-zA-Z\\d_$]*\\.)*[a-zA-Z_$][a-zA-Z\\d_$]*$');
   if (!re.test(userInput)) {
-    return 'Invalid groupId';
+    if (!/^[a-zA-Z_$]/.test(userInput)) {
+      return 'Invalid groupId: A valid groupId must start with a character from A to z, or one of the following symbols: _$';
+    } else if (!/[a-zA-Z\\d_$]$/.test(userInput)) {
+      return 'Invalid groupId: A valid groupId must end with a character from A to z, a number, or one of the following symbols: _$';
+    }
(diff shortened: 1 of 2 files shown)


src/generateProject/validateInput.ts
```diff
@@ -47,15 +51,23 @@ export async function validatePackageName(userInput: string): Promise<string|und
...
+    } else if (!/[a-zA-Z\\d_$]$/.test(userInput)) {
+      return 'Invalid package name: A valid package name must end with characters from A to z, a number, or the following symbols: _$';
+    }
+    return 'Invalid package name: A valid package name can only contain characters from A to z and the following symbols: ._$';
   }
   return undefined;
 }
 
 export async function validateResourceName(userInput: string): Promise<string|undefined> {
   const re = new RegExp('^[A-Za-z]+[A-Za-z0-9_.]*$');
```
fbricon: `.` is not a valid character for a resource name

src/generateProject/validateInput.ts
```diff
@@ -47,15 +51,23 @@ export async function validatePackageName(userInput: string): Promise<string|und
...
   if (!re.test(userInput)) {
-    return 'Invalid resource name';
+    if (!/^[A-Za-z]/.test(userInput)) {
+      return 'Invalid resource name: A valid resource name must start with a character from A to z';
+    }
+    return 'Invalid resource name: A valid resource name can only contain characters from A to z, numbers, and the following symbols: ._';
```
fbricon: `.` is not a valid character for a resource name

fbricon (changes requested):

src/generateProject/validateInput.ts
xorye: Did you mean package name?

fbricon: no, resource (aka class) name

xorye:
fixed

fbricon (approved):

Issue #112, opened 2019-10-16
Add quarkus extension to a Gradle project


fbricon:
Need to execute the gradle command to add a new extension

PR #117, opened 2019-10-17
Create launch config and debug task for Quarkus Gradle projects

### LINKED ISSUE #83: Generate launch config for Gradle based projects
See https://quarkus.io/guides/gradle-tooling

- need to detect project is based on gradle
- uses the quarkus plugin
- call a task executing `./gradlew quarkusDev`

xorye:
Fixes #83 
Tested on Windows
Sample Gradle-based Quarkus project: https://github.com/xorye/sample-quarkus-gradle-project/tree/master

### What this PR does
If the "Debug current Quarkus project" command is ran while a Quarkus Gradle project is open, a launch config and task will be added so that the `quarkusDev` command runs when a debug session starts. The code will use the Gradle wrapper file if it exists.


xorye:
I've created an interface called [IBuild](https://github.com/redhat-developer/vscode-quarkus/blob/eea5ec998cd07809b01b2abbbb620639b8a0c84e/src/definitions/IBuild.ts), where [MavenBuild](https://github.com/redhat-developer/vscode-quarkus/blob/eea5ec998cd07809b01b2abbbb620639b8a0c84e/src/definitions/MavenBuild.ts) and [GradleBuild](https://github.com/redhat-developer/vscode-quarkus/blob/eea5ec998cd07809b01b2abbbb620639b8a0c84e/src/definitions/GradleBuild.ts) implement it.

MavenBuild and GradleBuild instances are created here: https://github.com/redhat-developer/vscode-quarkus/blob/eea5ec998cd07809b01b2abbbb620639b8a0c84e/src/utils/workspaceUtils.ts#L38-L48

For #112 , I think it would be a good idea to add functions like `wrapperExists()`, `getDefaultExecutable()` to IBuild, and remove similar functions in [mavenUtils.ts](https://github.com/redhat-developer/vscode-quarkus/blob/eea5ec998cd07809b01b2abbbb620639b8a0c84e/src/utils/mavenUtils.ts) and [gradleUtils.ts](https://github.com/redhat-developer/vscode-quarkus/blob/eea5ec998cd07809b01b2abbbb620639b8a0c84e/src/utils/gradleUtils.ts).

angelozerr:
I test this PR with maven and gradle project (not in the same workspace) by starting several time the debug launch (to check the task was not created several time, remove the task at hand and restart it) and everything works very well.

src/utils/workspaceUtils.ts
```diff
@@ -14,19 +14,41 @@
...
+ * Determines whether the Quarkus project located in `workspaceFolder` is a Maven project
+ * or a Gradle project
+ * @param workspaceFolder
+ */
+export async function getQuarkusProjectBuildSupport(workspaceFolder: WorkspaceFolder): Promise<IBuildSupport | undefined> {
+  if ((await getFilePathsFromWorkspace(workspaceFolder, 'pom.xml')).length > 0) {
+    return MavenBuildSupport();
+  }
+
+  if ((await getFilePathsFromWorkspace(workspaceFolder, 'build.gradle')).length > 0) {
```
fbricon, line 41: build.gradle should have precedence

src/debugging/DebugConfigCreator.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Quarkus application in `workspaceFolder` when a debug session starts
+ */
+export class DebugConfigCreator {
+
+  public static debugConfigCreator: DebugConfigCreator;
+
+  private dotVSCodeDir: string;
+
+  public static async createFiles(workspaceFolder: WorkspaceFolder): Promise<void> {
+    DebugConfigCreator.debugConfigCreator = new DebugConfigCreator(workspaceFolder);
```
fbricon: avoid mutating static constants in static methods. You can prolly remove the static keyword from debugConfigCreator

src/debugging/LaunchConfigCreator.ts
```diff
@@ -0,0 +1,92 @@
...
+import { DebugConfiguration, WorkspaceFolder, workspace } from 'vscode';
+import { IBuildSupport } from '../definitions/IBuildSupport';
+import { FsUtils } from '../utils/fsUtils';
+/**
+ * This class is responsible for creating the debug Task that calls the
+ * Quarkus dev command.
+ */
+export class LaunchConfigCreator {
+
+  public static launchConfig: LaunchConfigCreator;
```
fbricon: don't keep it static

src/debugging/LaunchConfigCreator.ts
```diff
@@ -0,0 +1,92 @@
...
+  private quarkusBuildSupport: IBuildSupport;
+  private launchJsonFile: string;
+
+  /**
+   * Returns a debug task that calls the Quarkus dev command
+   * @param workspaceFolder the workspaceFolder containing the Quarkus project
+   * @param quarkusBuildSupport specifies whether the project in `workspaceFolder` is a Maven project or Gradle project
+   */
+  public static async createLaunchConfig(workspaceFolder: WorkspaceFolder, quarkusBuildSupport: IBuildSupport): Promise<void> {
+    LaunchConfigCreator.launchConfig = new LaunchConfigCreator(workspaceFolder, quarkusBuildSupport);
```
fbricon: keep as a local variable

fbricon (changes requested):

fbricon:
1st time I try to use the debug quarkus command, if no launch/task exist, I'm greeted with:

Screen Shot 2019-10-28 at 6 20 01 PM

src/debugging/DebugConfigCreator.ts
xorye: > avoid mutating static constants in static methods

Is this a rule/convention that is commonly followed?

fbricon: if 2 threads start mutating the same static object, you're in trouble

xorye:
@fbricon I can't reproduce that problem for both maven and gradle projects, the command creates the task and launch config and runs the debug session on my end.

I've tried it with the gradle project I mentioned in my original post and a new maven project generated using the generate project command.

Could you explain further what you did?

fbricon:
just cloned your test project, opened it, executed the debug quarkus command

xorye:
I've just tried it and it works for both linux and windows. I can try on mac tomorrow.

We're supposed to open the project using 'Open Folder' right? Click 'Open Folder' and choose 'sample-quarkus-gradle-project'?

fbricon:
yes

xorye:
did the task and launch config get created?

xorye:
I've added functions that specifically wait until the task & and launch config to be created. I wait for the task here:
https://github.com/xorye/vscode-quarkus/blob/0f2be545f5164fbd2927bfa5c1712f2b20c0350d/src/debugging/TaskCreator.ts#L42

and I wait for the launch config here:
https://github.com/xorye/vscode-quarkus/blob/0f2be545f5164fbd2927bfa5c1712f2b20c0350d/src/debugging/LaunchConfigCreator.ts#L42

To test this out on my machine, I added a set timeout to delay creating the task and launch config. The functions above successfully waits until the task and launch config are created (at least for my machine). Of course, I removed the set timeouts after.

@fbricon do you still get the error? 

I've come across a new problem:
1. Debug a new quarkus project by calling the command (task and launch config created successfully)
2. Exit the debug session, terminate the quarkus dev task
3. Delete the .vscode folder
4. Quickly call the debug command again
5. Could not find task 'quakusDev' error should appear

The error appears because after the task is deleted by removing the .vscode folder, the [vscode.tasks.fetchTasks()](https://github.com/microsoft/vscode/blob/557cf012ca0d1e25366af58185de5221d47b39a3/src/vs/vscode.d.ts#L5656) function still returns the deleted task.

I have another idea:
Instead of creating a tasks.json and launch.json, create only a launch.json that looks something like this:
```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Debug Quarkus application",
      "type": "java",
      "request": "attach",
      "port": 5005,
      "preLaunchTask": "shell: quarkusdev"
    }
  ]
}
```

and in the vscode-quarkus, we could register a Task Provider that provides a quarkusdev task:
```
export class FuncTaskProvider implements TaskProvider {
    public async provideTasks(token?: CancellationToken | undefined): Promise<Task[]> {
        const taskDefinition: TaskDefinition = { type: 'shell', name: 'quarkusdev' };
        const task: Task = (create quarkus dev task according to user's OS, gradle/maven, wrapper/nowrapper)
        );
        return [task];
    }
 
    public async resolveTask(task: Task, token?: CancellationToken | undefined): Promise<Task> {
    }
}

vscode.tasks.registerTaskProvider('quarkusdev', new FuncTaskProvider);
```

In this approach, the preLaunchTask could be generated every time the user runs the debug quarkus project command. This could be good because here, we can check the user's OS, build type, wrapper to come up with an appropriate command, instead of looking at a predefined task in tasks.json.

Also, this approach could circumvent problems with tasks being not found, tasks not being created in time, that new problem I just mentioned earlier etc, because we are not looking at tasks.json anymore. 

One downside to this is that the user would need to have vscode-quarkus installed to get that debug task (since tasks.json wouldn't contain the debug task anymore).

I strongly think that every feature would stay the same (debug terminate message would still work) but to be 100% sure, I would need to investigate more.

@fbricon what do you think about this idea?

fbricon:
Thanks @xorye !

PR #123, opened 2019-10-22
Change "Quarkus: Generate a Maven Project" command ID

xorye:
Changes the command id for creating Quarkus projects.
The term "Maven" was removed from the id because the command would also allow the users to create Gradle projects in the future.

Also, this change would be nice because adding the "Create a Quarkus Project" link for the Java extension pack overview page: https://github.com/Microsoft/vscode-java-pack, requires the vscode-quarkus command id for creating Quarkus projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "starter"
   ],
   "activationEvents": [
-    "onCommand:quarkusTools.createMavenProject",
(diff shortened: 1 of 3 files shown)


fbricon:
also need to update the welcome page with the proper command id

xorye:
fixed

PR #128, opened 2019-10-28
Display welcome page on install

### LINKED ISSUE #127: Welcome page appears too late
Currently, the welcome page appears when one of the activation events in [package.json](https://github.com/redhat-developer/vscode-quarkus/blob/17b9f08f19c6fcac7d8bbfb714e508974aa2e18a/package.json#L34-L38) are triggered.

However, in this scenario, the welcome page seems really out of place:
A user opens a Quarkus project and doesn't open application.properties
The user works on a java file for 30 minutes
The user opens application.properties and sees the welcome page

Seems weird to see a Quarkus welcome page after 30 minutes of working on a Quarkus project

xorye:
Fixes #127 

This PR implements a creative/hacky fix to display the welcome page on extension install. (ie, displays the welcome page when vscode-quarkus is installed)

The fix was to have vscode-quarkus activate on startup, show the welcome page and have vscode-quarkus manually edit the package.json file so that the startup activation event is removed.

Hi @jrieken, what do you think about this fix?

I have created a feature request for an "onExtensionInstalled" activation event to the VS Code GitHub page here: https://github.com/microsoft/vscode/issues/83451

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "onCommand:quarkusTools.addExtension",
     "onCommand:quarkusTools.debugQuarkusProject",
     "onCommand:quarkusTools.welcome",
-    "onLanguage:quarkus-properties"
+    "onLanguage:quarkus-properties",
+    "*"
   ],
   "main": "./dist/extension",
   "extensionDependencies": [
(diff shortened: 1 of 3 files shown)


jrieken:
> Hi @jrieken, what do you think about this fix?

I think that's a bad idea and you might trigger an "installation is corrupt" warning when changing your extension after install. Generally, we are looking into a streamlined way that allows extensions to show a welcome or getting started page and my recommendation is to join us on designing that. /cc @eamodio

fbricon:
damn that was a cool trick ;-)

@jrieken is there an existing issue we could start watching about improving the 3rd party welcome page situation?

PR #131, opened 2019-10-30
Add new headers for code.quarkus.io requests

### LINKED ISSUE #130: Add new header parameters when calling code.quarkus.io
```
API consumers shoult add two new header params:
Client-Name: the name of the client (ex: "Code Quarkus")
Client-Contact-Email: the email to contact in case of change in the API
```

Client-Name: vscode-quarkus
Client-Contact-Email: fbricon (use redhat email)

@ia3andy should we add the version in the client name? Can we send multiple contact headers?

xorye:
Fixes #130 

### DIFF
--- a/src/utils/requestUtils.ts
+++ b/src/utils/requestUtils.ts
@@ -25,6 +25,11 @@ import { Readable } from 'stream';
 import { ZipFile, fromBuffer } from 'yauzl';
 import { convertToQExtension } from '../definitions/QExtension';
 
+const HEADERS = {
+  'Client-Name': 'vscode-quarkus',
+  'Client-Contact-Email': 'fbricon@redhat.com,azerr@redhat.com,dakwon@redhat.com'
+};
+
 export async function getQExtensions(): Promise<QExtension[]> {
   const apiUrl: string = QuarkusConfig.getApiUrl();
   const extensions: string = await tryGetExtensionsJSON(apiUrl);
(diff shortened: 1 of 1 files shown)


src/utils/requestUtils.ts
```diff
@@ -78,8 +85,16 @@ export async function downloadProject(state: ProjectGenState): Promise<ZipFile>
 }
 
 async function tryGetProjectBuffer(projectUrl: string): Promise<Buffer> {
+  const requestOptions: request.OptionsWithUri = {
+    uri: projectUrl,
+    headers: {
+      'Client-Name': CLIENT_NAME,
+      'Client-Contact-Email': CLIENT_CONTACT_EMAIL
```
fbricon: is it possible to define a headers constant holding those values? to avoid repeating it for each query

xorye: I couldnt find anything in the docs: https://www.npmjs.com/package/request#custom-http-headers

Should I put the `request()` in a custom wrapper class to allow that?

fbricon: simply do 
```ts
const HEADERS = {
  'Client-Name': 'vscode-quarkus',
  'Client-Contact-Email': 'fbricon@redhat.com,azerr@redhat.com,dakwon@redhat.com'
};
...

const requestOptions: request.OptionsWithUri = {
    uri: `${apiUrl}/extensions`,
    headers: HEADERS,
    timeout: 30000
  };
...
```

xorye: thanks, I've pushed the change

PR #145, opened 2019-11-08
Use language instead of pattern for documentSelector

angelozerr:
Use language instead of pattern for documentSelector

As 'quarkus-properties' is defined as language in package.json, I think it's better to use it for documentSelector.


fbricon:
LGTM but I'd prefer if @xorye could test it

xorye:
Works for me

xorye (approved):

PR #149, opened 2019-11-12
Fix uri comparison

tolusha:
Signed-off-by: Anatoliy Bazko <abazko@redhat.com>


fbricon:
Can you please add a description of the issue in the PR, so that we have some context

tolusha:
I tried to run the extension in the Theia IDE and figured out that extension can't find a task that had been generated. Some how it works in VS Code but in Theia `TaskWorkpace` and `WorkspaceFolder` contain different instances of the same URI. So, we have to compare them using `toString()` function.

fbricon:
Thanks @tolusha !

PR #150, opened 2019-11-12
Add disposables to context.subscriptions

xorye:
This PR adds disposables to `context.subscriptions` instead of a separate array that is manually disposed of in the `deactivate()` function.

The contents of `context.subscriptions` are automatically disposed of when the extension is deactivated. [Source](https://github.com/microsoft/vscode/blob/12b7d0aafeaaa65e683e9289126bc75db2278c5c/src/vs/vscode.d.ts#L4941-L4945)


xorye:
And notably, this PR deals with the `debug.onDidTerminateDebugSession` disposable properly.

Instead of doing: 
`disposables.push(debug.onDidTerminateDebugSession(() => {}, null, disposables));`

this PR does:
`context.subscriptions.push(debug.onDidTerminateDebugSession(() => {}));`

PR #152, opened 2019-11-12
Register `quarkus.command.configuration.update` command

xorye:
Registers the `quarkus/clientSettingsUpdate` command for https://github.com/redhat-developer/quarkus-ls/pull/141

### DIFF
--- a/src/QuarkusConfig.ts
+++ b/src/QuarkusConfig.ts
@@ -16,7 +16,7 @@
 import {
   DEFAULT_API_URL,
 } from './definitions/constants';
-import { workspace } from 'vscode';
+import { ConfigurationTarget, workspace } from 'vscode';
 
 /**
  * This class manages the extension's interaction with
@@ -43,15 +43,15 @@ export namespace QuarkusConfig {
   }
 
   export function setAlwaysShowWelcomePage(value: boolean): void {
-    workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(ALWAYS_SHOW_WELCOME_PAGE), value, true);
+    saveToQuarkusConfig(ALWAYS_SHOW_WELCOME_PAGE, value);
   }
 
   export function setTerminateProcessOnDebugExit(value: string): void {
-    workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(DEBUG_TERMINATE_ON_EXIT), value, true);
+    saveToQuarkusConfig(DEBUG_TERMINATE_ON_EXIT, value);
   }
 
-  function removeQuarkusConfigName(configName: string) {
-    return configName.replace(QUARKUS_CONFIG_NAME + '.', '');
+  export function saveToQuarkusConfig<T>(configName: string, value: T) {
+    workspace.getConfiguration().update(configName, value, ConfigurationTarget.Global);
   }
 }
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -17,7 +17,7 @@ import * as requirements from './languageServer/requirements';
 
 import { JdtLSCommands, QuarkusLS, VSCodeCommands } from './definitions/constants';
 
-import { DidChangeConfigurationNotification, Disposable, LanguageClientOptions, LanguageClient, RequestType } from 'vscode-languageclient';
+import { DidChangeConfigurationNotification, LanguageClientOptions, LanguageClient, RequestType } from 'vscode-languageclient';
 import { ExtensionContext, commands, window, workspace } from 'vscode';
 import { QuarkusContext } from './QuarkusContext';
 import { addExtensionsWizard } from './addExtensions/addExtensionsWizard';
@@ -27,6 +27,7 @@ import { prepareExecutable } from './languageServer/javaServerStarter';
 import { tryStartDebugging } from './debugging/startDebugging';
 import { WelcomeWebview } from './webviews/WelcomeWebview';
 import { QuarkusConfig } from './QuarkusConfig';
+import { registerConfigurationUpdateCommand } from './lsp-commands';
 
 let languageClient: LanguageClient;
 
@@ -53,8 +54,8 @@ export function activate(context: ExtensionContext) {
   context.subscriptions.push(createTerminateDebugListener());
 
   connectToLS(context).then(() => {
-    const quarkusPojectInfoRequest = new RequestType<QuarkusProjectInfoParams, any, void, void>(QuarkusLS.PROJECT_REQUEST);
-    languageClient.onRequest(quarkusPojectInfoRequest, async (params: QuarkusProjectInfoParams) =>
+    const quarkusProjectInfoRequest = new RequestType<QuarkusProjectInfoParams, any, void, void>(QuarkusLS.PROJECT_REQUEST);
+    languageClient.onRequest(quarkusProjectInfoRequest, async (params: QuarkusProjectInfoParams) =>
        <any> await commands.executeCommand("java.execute.workspaceCommand", JdtLSCommands.PROJECT_INFO_COMMAND, params)
     );
 
@@ -70,6 +71,10 @@ export function activate(context: ExtensionContext) {
       languageClient.sendNotification("quarkus/quarkusPropertiesChanged", event);
     }));
 
+    /**
+     * Register standard LSP commands
+     */
+    context.subscriptions.push(registerConfigurationUpdateCommand());
   }).catch((error) => {
     window.showErrorMessage(error.message, error.label).then((selection) => {
       if (error.label && error.label === selection && error.openUrl) {
@@ -129,7 +134,16 @@ function connectToLS(context: ExtensionContext) {
       ],
       // wrap with key 'settings' so it can be handled same a DidChangeConfiguration
       initializationOptions: {
-        settings: getQuarkusSettings()
+        settings: getQuarkusSettings(),
+        extendedClientCapabilities: {
+          commands: {
+            commandsKind: {
+                valueSet: [
+                    "quarkus.command.configuration.update"
+                ]
+            }
+        }
+        }
       },
       synchronize: {
         // preferences starting with these will trigger didChangeConfiguration
--- a/src/lsp-commands.ts
+++ b/src/lsp-commands.ts
@@ -1,0 +1,93 @@
+/**
+ * Copyright 2019 Red Hat, Inc. and others.
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *     http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ConfigurationTarget, Disposable, commands, workspace } from 'vscode';
+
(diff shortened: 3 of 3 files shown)


src/QuarkusConfig.ts
```diff
@@ -49,6 +50,12 @@ export namespace QuarkusConfig {
   export function setTerminateProcessOnDebugExit(value: string): void {
     workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(DEBUG_TERMINATE_ON_EXIT), value, true);
   }
+  // "quarkus.tools.validation.unknown.excluded
+  export function addToUnknownExcluded(value: string): void {
+    const excluded: string[] = workspace.getConfiguration().get<string[]>(UNKNOWN_EXCLUDED, []);
+    excluded.push(value);
```
fbricon: check if it doesn't already exist in the list

src/extension.ts
```diff
@@ -70,6 +70,12 @@ export function activate(context: ExtensionContext) {
       languageClient.sendNotification("quarkus/quarkusPropertiesChanged", event);
     }));
 
+    context.subscriptions.push(commands.registerCommand('quarkus/clientSettingsUpdate', (setting: string, value: any) => {
+      if (setting === 'validationUnknownExcluded') {
```
fbricon: full property key (quarkus.tools.validation.unknown.excluded) should be received thus checked

src/extension.ts
```diff
@@ -70,6 +70,12 @@ export function activate(context: ExtensionContext) {
       languageClient.sendNotification("quarkus/quarkusPropertiesChanged", event);
     }));
 
+    context.subscriptions.push(commands.registerCommand('quarkus/clientSettingsUpdate', (setting: string, value: any) => {
```
fbricon: there should be a flag to indicate if the preference should be added, updated, deleted

src/extension.ts
```diff
@@ -70,6 +70,12 @@ export function activate(context: ExtensionContext) {
...
+      if (setting === 'validationUnknownExcluded') {
+        QuarkusConfig.addToUnknownExcluded(value as string);
```
fbricon: QuarkusConfig.updatePreference(key, changeType, value);

fbricon (changes requested):

xorye:
In my latest commit, I've changed the quarkus-ls (see https://github.com/redhat-developer/quarkus-ls/pull/141) so that quarkus-ls sends an [ConfigurationItemEdit](https://github.com/rdhat-developer/quarkus-ls/blob/820a2d94fa81f4cf8492bae9d10c42421d3cb486/quarkus.ls/com.redhat.quarkus.ls/src/main/java/com/redhat/quarkus/services/ConfigurationItemEdit.java) object to the client (vscode-quarkus) which the client would read and make config changes.

I don't deal with config scope for now. I will look into it

angelozerr:
@xorye as we would like to propose our idea to LSP with command, I think we should have a more generic code for updating the configuration.

Keep in mind that LSP command capability should provides 4 "standard" commands:

 * open URI
 * references
 * implementations
 * update configuration.

The idea is to create a `lsp-commands.ts`  which could be copied/pasted in an another project (like vscode-xml) without changes (or few). Please create this file and move your whole code which register the commands of update configuration in this file. We will add other commands after that. I mean https://github.com/redhat-developer/vscode-xml/blob/6a7db6eb6eff235ef80f5fe9b0dd0ce33c9f5fc2/src/extension.ts#L99

This  `lsp-commands.ts` could provides a `function registerConfigurationUpdateCommand(section?: string): Disposable {` which register the `quarkus.command.configuration.update`

src/QuarkusConfig.ts
```diff
@@ -43,11 +46,41 @@ export namespace QuarkusConfig {
...
-    workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(ALWAYS_SHOW_WELCOME_PAGE), value, true);
+    saveToQuarkusConfig(ALWAYS_SHOW_WELCOME_PAGE, value);
   }
 
   export function setTerminateProcessOnDebugExit(value: string): void {
-    workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(DEBUG_TERMINATE_ON_EXIT), value, true);
+    saveToQuarkusConfig(DEBUG_TERMINATE_ON_EXIT, value);
+  }
+
+  export function resolveConfigurationItemEdit<T>(configItemEdit: ConfigurationItemEdit) {
```
angelozerr: Move all code with configuration update inside registerConfigurationUpdateCommand

src/lsp-commands.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Disposable, commands } from 'vscode';
+
+namespace CommandKind {
+  export const COMMAND_REFERENCES = "quarkus.command.references";
+  export const COMMAND_IMPLEMENTATIONS = "quarkus.command.implementations";
+  export const COMMAND_OPEN_URI = "quarkus.command.open.uri";
+  export const COMMAND_CONFIGURATION_UPDATE = "quarkus.command.configuration.update";
+}
+
+export function registerConfigurationUpdateCommand(callback: (...args: any[]) => any): Disposable {
```
angelozerr: remove callback parameter which should be implemented inside this registerConfigurationUpdateCommand function. If you need to add some parameter (like section, do it), but the idea is to copy/paste this file in an another vscode extension and it should work without changes (just change the `quarkus.command.configuration.update` into `xml.command.configuration.update` and that's all

src/lsp-commands.ts
```diff
@@ -0,0 +1,89 @@
...
+namespace CommandKind {
+  export const COMMAND_REFERENCES = "quarkus.command.references";
+  export const COMMAND_IMPLEMENTATIONS = "quarkus.command.implementations";
+  export const COMMAND_OPEN_URI = "quarkus.command.open.uri";
+  export const COMMAND_CONFIGURATION_UPDATE = "quarkus.command.configuration.update";
+}
+
+export function registerConfigurationUpdateCommand(): Disposable {
+  return commands.registerCommand(CommandKind.COMMAND_CONFIGURATION_UPDATE, (configItemEdit: ConfigurationItemEdit) => {
+    if (configItemEdit.section === QuarkusConfig.UNKNOWN_EXCLUDED) {
```
angelozerr: why you need this condition?

src/lsp-commands.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export function registerReferencesCommand(): Disposable {
+  return commands.registerCommand(CommandKind.COMMAND_REFERENCES, () => {
+
+  });
+}
+
+function resolveConfigurationItemEdit<T>(configItemEdit: ConfigurationItemEdit) {
+  if (!QuarkusConfig.ALL_CONFIG_SECTIONS.includes(configItemEdit.section)) {
```
angelozerr: Why you need this condition?

src/lsp-commands.ts
```diff
@@ -0,0 +1,85 @@
...
+namespace CommandKind {
+  export const COMMAND_REFERENCES = "quarkus.command.references";
+  export const COMMAND_IMPLEMENTATIONS = "quarkus.command.implementations";
+  export const COMMAND_OPEN_URI = "quarkus.command.open.uri";
+  export const COMMAND_CONFIGURATION_UPDATE = "quarkus.command.configuration.update";
+}
+
+export function registerConfigurationUpdateCommand(): Disposable {
+  return commands.registerCommand(CommandKind.COMMAND_CONFIGURATION_UPDATE, (configItemEdit: ConfigurationItemEdit) => {
+    if (configItemEdit.section === 'quarkus.tools.validation.unknown.excluded') {
```
angelozerr: do we need this condition?

src/lsp-commands.ts
```diff
@@ -0,0 +1,85 @@
...
+import { ConfigurationTarget, Disposable, commands, workspace } from 'vscode';
+
+namespace CommandKind {
+  export const COMMAND_REFERENCES = "quarkus.command.references";
+  export const COMMAND_IMPLEMENTATIONS = "quarkus.command.implementations";
+  export const COMMAND_OPEN_URI = "quarkus.command.open.uri";
+  export const COMMAND_CONFIGURATION_UPDATE = "quarkus.command.configuration.update";
+}
+
+export function registerConfigurationUpdateCommand(): Disposable {
```
angelozerr, line 29: Add documentation

src/extension.ts
```diff
@@ -70,6 +71,7 @@ export function activate(context: ExtensionContext) {
       languageClient.sendNotification("quarkus/quarkusPropertiesChanged", event);
     }));
 
+    context.subscriptions.push(registerConfigurationUpdateCommand());
```
angelozerr, line 77: Add comment like 'register standard LSP commands'

src/QuarkusConfig.ts
```diff
@@ -43,11 +44,15 @@ export namespace QuarkusConfig {
...
+    saveToQuarkusConfig(ALWAYS_SHOW_WELCOME_PAGE, value);
   }
 
   export function setTerminateProcessOnDebugExit(value: string): void {
-    workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(DEBUG_TERMINATE_ON_EXIT), value, true);
+    saveToQuarkusConfig(DEBUG_TERMINATE_ON_EXIT, value);
+  }
+
+  export function saveToQuarkusConfig<T>(configName: string, value: T) {
+    workspace.getConfiguration(QUARKUS_CONFIG_NAME).update(removeQuarkusConfigName(configName), value, true);
```
angelozerr: Could be not replaced with 

```ts
workspace.getConfiguration().update(configName, value, ConfigurationTarget.Workspace);
```

src/QuarkusConfig.ts
```diff
@@ -29,6 +29,7 @@ export namespace QuarkusConfig {
   export const STARTER_API = 'quarkus.tools.starter.api';
   export const ALWAYS_SHOW_WELCOME_PAGE = 'quarkus.tools.alwaysShowWelcomePage';
   export const DEBUG_TERMINATE_ON_EXIT = 'quarkus.tools.debug.terminateProcessOnExit';
+  export const UNKNOWN_EXCLUDED = 'quarkus.tools.validation.unknown.excluded';
```
angelozerr: remove this dead code.

src/lsp-commands.ts
xorye: the reason I have this condition is because of this feedback from a while ago: https://github.com/redhat-developer/vscode-quarkus/pull/152#discussion_r345753500

do you think we don't need it?

src/QuarkusConfig.ts
xorye: I'm not sure if we want to do `ConfigurationTarget.Workspace` here, because that would mean the `alwaysShowWelcomePage` would depend on Workspace settings when I think it should be User/Global settings instead.

I think `quarkus.tools.debug.terminateProcessOnExit` can be Workspace but I think `quarkus.tools.starter.api` and `quarkus.tools.alwaysShowWelcomePage` should be Global/User. @fbricon what do you think?

src/lsp-commands.ts
angelozerr: I would prefer remove it to have no dependency to Quarkus. If we find issues, we will fix it by adding a parameter for instance. But please keep this code with  no dependencies to Quarkus.

src/QuarkusConfig.ts
angelozerr: @xorye keep as it worked. Change the ConfigurationTarget to have the same code. But IMHO I think it's more clean to use workspace.getConfiguration().update.

xorye: Ok great. I agree, `workspace.getConfiguration().update` is much cleaner

xorye:
I have updated it

PR #146, opened 2019-11-12
Provide codelenses for endpoints of the running app

angelozerr:
Provide codelenses for endpoints of the running app

See https://github.com/redhat-developer/quarkus-ls/issues/114

Signed-off-by: azerr <azerr@redhat.com>

PR #163, opened 2019-11-25
Provide hover for ConfigProperty name

xorye:
The branch for this PR has been branched off of https://github.com/angelozerr/vscode-quarkus/tree/codelens, so it would be nice if https://github.com/redhat-developer/vscode-quarkus/pull/146 is merged before this PR is merged.

This PR works alongside:

### DIFF
--- a/src/definitions/constants.ts
+++ b/src/definitions/constants.ts
@@ -19,12 +19,14 @@ export namespace QuarkusLS {
   export const PROJECT_REQUEST = 'quarkus/projectInfo';
   export const PROPERTY_DEFINITION_REQUEST = 'quarkus/propertyDefinition';
   export const JAVA_CODELENS_REQUEST = 'quarkus/java/codeLens';
+  export const JAVA_HOVER_REQUEST = 'quarkus/java/hover';
 }
 
 export namespace JdtLSCommands {
   export const PROJECT_INFO_COMMAND = 'quarkus.java.projectInfo';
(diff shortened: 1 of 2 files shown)


src/definitions/constants.ts
```diff
@@ -33,6 +35,7 @@ export namespace VSCodeCommands {
   export const ADD_EXTENSIONS = 'quarkusTools.addExtension';
   export const DEBUG_QUARKUS_PROJECT = 'quarkusTools.debugQuarkusProject';
   export const QUARKUS_WELCOME = 'quarkusTools.welcome';
+  export const QUARKUS_OPEN_URL = 'quarkus.open.url';
```
angelozerr: Why you need that?

src/extension.ts
```diff
@@ -130,6 +140,13 @@ function registerVSCodeCommands(context: ExtensionContext) {
     WelcomeWebview.createOrShow(context);
   }));
 
+  /**
+   * Command for opening CodeLens URL
+   */
+  context.subscriptions.push(commands.registerCommand(VSCodeCommands.QUARKUS_OPEN_URL, (url) => {
```
angelozerr: Why you need that?

xorye: Looks like it is some old code from before, I will remove it

src/definitions/constants.ts
xorye: Looks like it is some old code from before, I will remove it

xorye:
@angelozerr fixed

angelozerr:
Thanks @xorye !

PR #168, opened 2019-12-02
Ignore images/

xorye:
Ignores the `images/` directory when packaging the extension with `vsce package`.
https://code.visualstudio.com/api/working-with-extensions/publishing-extension#.vscodeignore

The resulting vsix file is about 0.5mb smaller.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -5,5 +5,6 @@
 node_modules/
 src/
 out/
+images/
 tsconfig.json
 webpack.config.js


fbricon:
Once you install the visx, how does the readme look like in the extensions view?

xorye:
We still see the main image:
image

This is because `vsce package` will replace the relative links in the readme to absolute urls.
When I open the .vsix file and look at the readme, it looks like this:
```
![](https://github.com/redhat-developer/vscode-quarkus/raw/master/images/propertiesSupport.png)
```

`vsce package` does this even if images/ is not ignored

PR #169, opened 2019-12-03
Add different ways to open new generated project

### LINKED ISSUE #157: Provide option to add created new project folder to multi-root VS Code workspace
Currently when the user activates the "Quarkus: Generate a Maven Project" command, the result is a whole new workbench window with the new project folder as the root. I suspect this is due to the old single folder workspace in VS Code. Now that VS Code has multi-root workspaces (https://code.visualstudio.com/docs/editor/multi-root-workspaces) it would be much less jarring to the user to have the new folder simply added to the workspace rather than opening an entirely separate window. Or at least it would be good to have the option to go either way.

xorye:
Fixes #157

Generating a new project with a workspace or folder open: 
image

Generating a new project with nothing open:
image

### DIFF
--- a/src/generateProject/generationWizard.ts
+++ b/src/generateProject/generationWizard.ts
@@ -9,7 +9,7 @@ import * as fse from 'fs-extra';
 
 import { INPUT_TITLE, BuildToolName } from '../definitions/constants';
 import { MultiStepInput } from '../utils/multiStepUtils';
-import { OpenDialogOptions, QuickPickItem, Uri, commands, window } from 'vscode';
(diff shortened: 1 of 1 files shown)


xorye:
Tested on Windows 10 and Fedora 30

fbricon:
Generating a new project with nothing open: should directly open in the current window, no dialog necessary

fbricon:
I'm getting this error, when running the generate project command on a new window with no open folders: 
Screen Shot 2019-12-05 at 3 26 30 PM

xorye:
Here is the issue I created about that^

https://github.com/microsoft/vscode/issues/86419

fbricon:
Nice work @xorye!

PR #180, opened 2019-12-26
Use API to check if tasks.json and launch.json exist

svor:
Signed-off-by: svor <vsvydenk@redhat.com>
This PR suggests changes to use VS Code API to check if the workspace already has places (launch.json and tasks.json) to add new debug configuration and new task.

It could be useful if these configuration files are not located in .vscode folder. In case of Theia, they are in .theia folder

Which problem I've tried to solve: https://github.com/eclipse/che/issues/15523

@xorye @angelozerr @fbricon  Could you please take a look


xorye:
Thank you @svor , this PR looks good and works well on Fedora and Windows. @fbricon I'd say this can be merged.

PR #186, opened 2020-01-08
Qute textmate grammar

### LINKED ISSUE #182: Add Qute TextMate grammar and language configuration for HTML files
Create a `qute.tmLanguage` which includes html TextMate like https://github.com/angelozerr/lsp4e-freemarker/blob/master/org.eclipse.lsp4e.freemarker/syntaxes/ftl.tmLanguage
You can see a demo in this repository with HTML completion (coming from the HTML Language Server of vscode and Freemarker Language Server for validation).

Perhaps a good idea is to use those file extension:

*.html.qute (or if it doesn't work .htmlQute)
*.json.qute (or if it doesn't work .jsonQute)

This language configuration could be used in Eclipse IDE too with TM4E  /cc @jeffmaury  (and I think IJ supports too TextMate)

xorye:
Fixes #182 

To test this PR open a `.qute` file. 
This PR adds a `test.qute` file for you to try out.

## What this PR does

This PR adds a new Qute language with its own textmate grammar:
image

## Things to improve:
1. Within an if section, if we have a comparison with an integer or float, the integer should be a different colour:
image
```
{#if item.age > 10}
```

2. The same applies for strings:
image
```
{#if item.name is 'sword'}
```

3. If a method is being called, the method should be a different colour:
image
```
{#for item in item.getDerivedItems()} 
```

4. If brackets are being used inside of a string like in the following image, `{person.name}` should be a different colour.
image

```
First name: <input type="text" name="fname" value="{person.name}">
```

5. Currently, the colour scheme is what it is because the textmate grammar provided in this PR is based on FreeMarker's grammar: https://github.com/bburbach/textmate-freemarker-bundle/blob/master/FreeMarker.tmbundle/Syntaxes/FreeMarker.tmLanguage

The colours for keywords like "for" and "if" are given a root group of "entity", therefore its colour is yellow. I think it would be nice to be consistent with programming languages' syntaxes, I think it would be better if these keywords were given "keyword.control.(something)" instead.

Doing so, will give these keywords a more familiar colour (at least for VS Code): 
image

In VS Code, keywords like "for" and "if" from programming languages like Java, Python, Typescript are purple.

I'm not sure if this behaviour will be the same in IDEs like Eclipse or IntelliJ, so for now, this is just a suggestion. 

## Things I am not sure about
1. Should parameter declarations (`{@org.acme.Foo foo}` in the following image) be highlighted in a different colour?
image

2. The textmate grammar inherits rules from this HTML grammar: https://github.com/textmate/html.tmbundle/blob/master/Syntaxes/HTML.plist

So, therefore, HTML comments are highlighted in green:
image

From what I understand, Qute templates have their own comment style:
image

Are HTML comments valid in Qute templates?


angelozerr:
@xorye it looks like very promising! Congrats!

> To test this PR open a `.qute` file.

It's a thing that we must discuss with @mkouba because Qute documentation uses `.html` file extension and not `.qute`. 

We have 2 choices, @mkouba, @maxandersen  (and Qute community) : 

 * are OK to use `.qute` file extension. Your PR is perfect. But perhaps we should use `.html-qute` file extension because Qute can generate anything. So we could imagine to generate json content (`.json-qute`). In other words we could have a convention name for Qute template. What do you think about that?

 * wishes to keep `.html` file extension, is it possible to manage that with vscode, Eclipse IDE, etc? Perhaps Textmate injection could help us https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#creating-a-basic-injection-grammar but my fear is that it will update all html file from all project of vscode (not only project which have Qute dependency.

> Things to improve:

Indeed it should be very nice, but if it's too complex, perhaps it can be enough for the moment. @mkouba, @maxandersen, @fbricon what do you think about that?

> Should parameter declarations ({@org.acme.Foo foo} in the following image) be highlighted in a different colour?

I think it can be enough but I let @mkouba, @maxandersen, @fbricon give us feedback.

> Are HTML comments valid in Qute templates?

Yes

mkouba:
So I think that we should avoid using `.qute` if possible because:

* Qute itself is not a file format
* In Quarkus we use the file extension to detect the media type when working with variant templates

Maybe just use `.qute.html`? Otherwise we would have to parse the content of the file and look for the Qute constructs `{@ ...}`, `{! ... !}`, `{#...}`, etc.

angelozerr:
> So I think that we should avoid using `.qute`

It was my fear :) 

@xorye do you know if we could use  `.qute.html` as file type in vscode ?

maxandersen:
.qute.html is fine imo.

angelozerr:
> .qute.html is fine imo.

I agree but I don't know if vscode, Eclipse IDE, etc considers that `.qute.html` is a file extension. I fear that `.html` will be the file extension. We need to do some tests about that.

xorye:
I added a new commit that uses .qute.html and it seems to be working for VS Code:
image

angelozerr:
> I added a new commit that uses .qute.html and it seems to be working for VS Code:

Great! I will check that in Eclipse IDE.

fbricon:
mmm qute.html works as it's recognized as a qute file, but then we lose all the HTML support (completion, emmet ...)

maxandersen:
> mmm qute.html works as it's recognized as a qute file, but then we lose all the HTML support (completion, emmet ...)

how was that different when the file is called .html-qute or something else ?

angelozerr:
@maxandersen, @xorye  is studying how to manage this html file extension, we will give you feedback when we will find a clean solution, because the challenge is to have a file extension (if we could keep `.html` it should be really cool) which:

 * keeps the HTML support managed by the HTML LS (completion, hover, etc)
 * keeps the HTML syntax coloration and add Qute syntax (perhaps an injection of TextMate could work ?)
 * augments the HTML language configuration (matching bracket,etc) with Qute syntax (ex: {! !}).

angelozerr:
@maxandersen after several experimentation (correct me if I'm wrong @xorye ) it seems the better solution to support Qute syntax coloration with TextMate WITHOUT loosing HTML support (completion, etc) is NOT to define a Qute language. In other words the only solution is to inject Qute syntax with TextMate to the HTML TextMate grammar. 

It works great except that all HTML files have Qute syntax coloration even if a project is not a Qute project. So the idea is to give the capability to use to enable/disable the Qute syntax for a workspace. To do that an idea, @fbricon suggest to create a new vscode extension (ex : `vscode-qute-syntax-highlighting`) which injects only the Qute TextMate grammar in the given HTML TextMate grammar.

The user can disable the Qute syntax coloration by uninstalling the `vscode-qute-syntax-highlighting` and by keeping Quarkus (and Qute support managed by language server).

For Qute language configuration (to manage matching bracket for Qute comment ({! !}) I don't know how to do that?

If @fbricon is OK, @xorye please create this new vscode extension, thanks!

maxandersen:
that’s a bummer and funky limitation of vscode ;/

I reckon we'll still be able to have a generic ".qute" file support which will just syntax highlight the qute part ? 

how about content assist ? is that limited in the same way  or is this truly just for syntax highlighting things gets a bit convoluted ?

angelozerr:
> that’s a bummer and funky limitation of vscode ;/

Yes I know, you can read an interesting discussion at https://github.com/microsoft/vscode/issues/55088#issuecomment-408031810

> I reckon we'll still be able to have a generic ".qute" file support which will just syntax highlight the qute part ?

We could do that, but it will require that we will create a vcode `Qute language type` and we will loose HTML support completion, etc) because HTML language server works only with vscode `HTML language`

xorye:
@angelozerr @maxandersen 

It looks like VS Code could soon provide a contribution point for other html-like languages to register to the html language server, as mentioned in point b. in this comment: https://github.com/microsoft/vscode/issues/88377#issuecomment-573568651

xorye:
I pushed some improvements to the grammar:
image

There are now two grammar files: `qute-html.tmLanguage.json` and `qute.tmLanguage.json`.

### `qute-html.tmLanguage.json`
This is the grammar that the Qute HTML language uses.
It brings in the Qute grammar from `qute.tmLanguage.json`, brings in HTML grammar, and injects some parts of the Qute grammar into the HTML grammar.

###  `qute.tmLanguage.json`
This is the grammar file containing Qute specific grammar. This file does not contain anything specific to HTML.

### Improvements:
image

image

image

image

image

image

fbricon:
That looks pretty good @xorye, nice job!
~~However, I found a case where the highlighting is missing, when the elvis operator (?:)  is involved~~(edit: Nope, everything is fine)

angelozerr:
Awesome! Well done @xorye !

Issue #177, opened 2019-12-19
Improve release process to include sources when publishing vsix


nickboldt:
Similar to discussions in...

* https://github.com/camel-tooling/vscode-wsdl2rest/issues/142 
* https://github.com/redhat-developer/vscode-java/issues/892
* https://github.com/redhat-developer/vscode-project-initializer/issues/40
* https://github.com/camel-tooling/vscode-camelk/issues/243
* https://github.com/windup/rhamt-vscode-extension/issues/263
* https://issues.redhat.com/browse/FUSETOOLS2-80

... we'd like that each release of this .vsix file to dl.jb.o be accompanied by a matching sources.tar.gz

I don't care if you push the .vsix to GH releases page, or tar up sources and push that to dl.jb.o, just that we have binary and sources in the same place.

Thanks!

This is a RH product security requirement in that we need to either build everything used in CodeReady Workspaces from scratch as part of our build, or else provide matching sources for upstream RH-built binaries.

fbricon:
@nickboldt this should have been fixed with #177. Is this ok for you: https://download.jboss.org/jbosstools/vscode/snapshots/vscode-quarkus/?C=M;O=D?

nickboldt:
yes, but I need a /stable/ release, not a /snapshots/ one. 

https://download.jboss.org/jbosstools/vscode/stable/vscode-quarkus/

@davidkwon tells me you're planning a 1.3.0 release next week?

fbricon:
Yes, but there's a good chance it will slip to the week after, as we found some annoying issue that need to be fixed 1st

nickboldt:
OK np. Before mid Feb is fine for me :)

nickboldt:
Ping! checking for new 1.3.0 in https://download.jboss.org/jbosstools/vscode/stable/vscode-quarkus/  but nothing yet. ETA still mid-early Feb?

nickboldt:
https://download.jboss.org/jbosstools/vscode/stable/vscode-quarkus/
  | vscode-quarkus-1.3.0-242.tgz | 06-Feb-2020 13:01 | 504K |  
  | vscode-quarkus-1.3.0-242.vsix | 06-Feb-2020 13:01 | 4.7M |  

Done!

PR #187, opened 2020-01-14
Publish sources #177

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,6 +51,7 @@ node('rhel7'){
         dir ('vscode-quarkus') {
             def packageJson = readJSON file: 'package.json'
             sh "vsce package -o ../vscode-quarkus-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
+            sh "npm pack && mv vscode-quarkus-${packageJson.version}.tgz ../vscode-quarkus-${packageJson.version}-${env.BUILD_NUMBER}.tgz"
         }
     }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -51,6 +51,7 @@ node('rhel7'){
         dir ('vscode-quarkus') {
             def packageJson = readJSON file: 'package.json'
             sh "vsce package -o ../vscode-quarkus-${packageJson.version}-${env.BUILD_NUMBER}.vsix"
+            sh "npm pack && mv vscode-quarkus-${packageJson.version}.tgz vscode-quarkus-${packageJson.version}-${env.BUILD_NUMBER}.tgz"
```
apupier: wondering why there is a ../ in previous line. Is it changing something here?

fbricon: this step is running inside the vscode-quarkus directory. It's putting the vsix file in the parent dir, so you'll need to do the same

apupier: Thanks for the explanation. I modified, to also move the generated tgz source files to the parent dir.

fbricon:
@apupier thanks for the PR. Before merging, I ran a jenkins build (\#221)  with those changes, the sources have been pushed to https://download.jboss.org/jbosstools/vscode/snapshots/vscode-quarkus/?C=M;O=D

PR #189, opened 2020-01-14
Add support for YAML configuration files

angelozerr:
Add support for YAML configuration files

See https://github.com/redhat-developer/quarkus-ls/issues/112

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "onCommand:quarkusTools.welcome",
     "workspaceContains:**/src/main/resources/application.properties",
     "workspaceContains:**src/main/resources/META-INF/microprofile-config.properties",
+    "workspaceContains:**/src/main/resources/application.yaml",
     "onLanguage:microprofile-properties",
     "onLanguage:quarkus-properties",
     "onLanguage:java",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,7 +15,7 @@
  */
 import * as requirements from './languageServer/requirements';
 
-import { VSCodeCommands } from './definitions/constants';
+import { VSCodeCommands, MicroProfileLS } from './definitions/constants';
 
 import { DidChangeConfigurationNotification, LanguageClientOptions, LanguageClient } from 'vscode-languageclient';
 import { ExtensionContext, commands, window, workspace } from 'vscode';
@@ -28,6 +28,7 @@ import { tryStartDebugging } from './debugging/startDebugging';
 import { WelcomeWebview } from './webviews/WelcomeWebview';
 import { QuarkusConfig } from './QuarkusConfig';
 import { registerConfigurationUpdateCommand, registerOpenURICommand, CommandKind } from './lsp-commands';
+import { registerYamlSchemaSupport, MicroProfilePropertiesChangeEvent } from './yaml/YamlSchema';
 
 let languageClient: LanguageClient;
 
@@ -37,21 +38,29 @@ export function activate(context: ExtensionContext) {
 
   context.subscriptions.push(createTerminateDebugListener());
 
+  /**
+   * Register Yaml Schema support to manage application.yaml
+   */
+  const yamlSchemaCache = registerYamlSchemaSupport();
+
   connectToLS(context).then(() => {
+    yamlSchemaCache.then(cache => { if (cache) { cache.languageClient = languageClient; } });
 
     /**
      * Delegate requests from MicroProfile LS to the Java JDT LS
      */
-    bindRequest('microprofile/projectInfo');
-    bindRequest('microprofile/propertyDefinition');
-    bindRequest('microprofile/java/codeLens');
-    bindRequest('microprofile/java/hover');
+    bindRequest(MicroProfileLS.PROJECT_INFO_REQUEST);
+    bindRequest(MicroProfileLS.PROPERTY_DEFINITION_REQUEST);
+    bindRequest(MicroProfileLS.JAVA_CODELENS_REQUEST);
+    bindRequest(MicroProfileLS.JAVA_HOVER_REQUEST);
 
     /**
      * Delegate notifications from Java JDT LS to the MicroProfile LS
      */
-    bindNotification('microprofile/propertiesChanged');
-
+    context.subscriptions.push(commands.registerCommand(MicroProfileLS.PROPERTIES_CHANGED_NOTIFICATION, (event: MicroProfilePropertiesChangeEvent) => {
+      languageClient.sendNotification(MicroProfileLS.PROPERTIES_CHANGED_NOTIFICATION, event);
+      yamlSchemaCache.then(cache => { if (cache) cache.evict(event); });
+    }));
   }).catch((error) => {
     window.showErrorMessage(error.message, error.label).then((selection) => {
       if (error.label && error.label === selection && error.openUrl) {
@@ -62,14 +71,8 @@ export function activate(context: ExtensionContext) {
 
   function bindRequest(request: string) {
     languageClient.onRequest(request, async (params: any) =>
-      <any> await commands.executeCommand("java.execute.workspaceCommand", request, params)
+      <any>await commands.executeCommand("java.execute.workspaceCommand", request, params)
     );
-  }
-
-  function bindNotification(notification: string) {
-    context.subscriptions.push(commands.registerCommand(notification, (event: any) => {
-      languageClient.sendNotification(notification, event);
-    }));
   }
 
   registerVSCodeCommands(context);
@@ -135,12 +138,12 @@ function connectToLS(context: ExtensionContext) {
         extendedClientCapabilities: {
           commands: {
             commandsKind: {
-                valueSet: [
-                  CommandKind.COMMAND_CONFIGURATION_UPDATE,
-                  CommandKind.COMMAND_OPEN_URI
-                ]
+              valueSet: [
+                CommandKind.COMMAND_CONFIGURATION_UPDATE,
+                CommandKind.COMMAND_OPEN_URI
+              ]
             }
-        }
+          }
         }
       },
       synchronize: {
@@ -183,7 +186,7 @@ function connectToLS(context: ExtensionContext) {
       quarkus = defaultValue;
     } else {
       const x = JSON.stringify(configQuarkus); // configQuarkus is not a JSON type
-      quarkus = { quarkus : JSON.parse(x)};
+      quarkus = { quarkus: JSON.parse(x) };
     }
     return quarkus;
   }
--- a/src/yaml/YamlSchema.ts
+++ b/src/yaml/YamlSchema.ts
@@ -1,0 +1,149 @@
+/**
+ * Copyright 2020 Red Hat, Inc. and others.
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *     http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as vscode from 'vscode';
+import * as semver from 'semver';
+import { VSCODE_YAML_EXTENSION_ID, MICROPROFILE_SCHEMA, MICROPROFILE_SCHEMA_PREFIX } from "./YamlConstants";
+import { Uri } from 'vscode';
+import { LanguageClient } from 'vscode-languageclient';
+import { MicroProfileLS } from '../definitions/constants';
+
+// The function signature exposed by vscode-yaml:
+// 1. the requestSchema api will be called by vscode-yaml extension to decide whether the schema can be handled by this
+// contributor, if it returns undefined, means it doesn't support this yaml file, vscode-yaml will ask other contributors
+// 2. the requestSchemaContent api  will give the parameter uri returned by the first api, and ask for the json content(after stringify) of
+// the schema
+declare type YamlSchemaContributor = (schema: string,
+  requestSchema: (resource: string) => string | undefined,
+  requestSchemaContent: (uri: string) => Promise<string>) => void;
+
+interface JsonSchemaForProjectInfo {
+  projectURI: string;
+  jsonSchema: string;
+}
+
+export interface MicroProfilePropertiesChangeEvent {
+  projectURIs: Array<String>;
+}
+
+// Yaml Schema cache which caches YAML Schema (in JSON format) for application.yaml files
+// which belong to a MicroProfile project.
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -36,9 +36,11 @@
     "onCommand:quarkusTools.debugQuarkusProject",
     "onCommand:quarkusTools.welcome",
     "workspaceContains:**/src/main/resources/application.properties",
+    "workspaceContains:**/src/main/resources/application.yaml",
     "onLanguage:quarkus-properties",
     "onLanguage:java",
-    "onLanguage:qute"
+    "onLanguage:qute",
+    "onLanguage:yaml"
```
fbricon: that's probably too eager. See if you can remove it

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,152 @@
...
+async function activateYamlExtension(): Promise<{ registerContributor: YamlSchemaContributor } | undefined> {
+  const ext = vscode.extensions.getExtension(VSCODE_YAML_EXTENSION_ID);
+  if (!ext) {
+    vscode.window.showWarningMessage('Please install \'YAML Support by Red Hat\' via the Extensions pane.');
+    return undefined;
+  }
+  const yamlPlugin = await ext.activate();
+
+  if (!yamlPlugin || !yamlPlugin.registerContributor) {
+    vscode.window.showWarningMessage('The installed Red Hat YAML extension doesn\'t support Kubernetes Intellisense. Please upgrade \'YAML Support by Red Hat\' via the Extensions pane.');
```
fbricon: Kubernetes? ;-)

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,152 @@
...
+    return undefined;
+  }
+
+  if (ext.packageJSON.version && !semver.gte(ext.packageJSON.version, '0.0.15')) {
```
fbricon, line 121: @tsmaeder if Che doesn't have the proper vscode-yaml version, will users be able to install a more recent version?

fbricon (changes requested):

angelozerr:
This PR requires the MicroProfile LS PR https://github.com/redhat-developer/quarkus-ls/pull/191 (this PR provides a new LSP service which convert the MicroProfileProjectInfo in a JSON Schema used by the YAML support). 

Please note this MicroProfile LS PR must be improved to support value type (string, number, array, etc). I will continue to work on it.

This PR uses the vscode-yaml, so you need to install vscode-yaml to play with it. The check of vscode-yaml installation is done with the same thing than https://github.com/Azure/vscode-kubernetes-tools but I can change anything if you wish.

As it uses the vscode-yaml, if you have Spring Boot vscod extension, please change the type of the file to YAML otherwise the yaml ls will never called.

The YAML support have some limitations: 

 * when you open the application.yaml at first, it will not work because vscode-yaml starts, checks if there is a registerContributor (to register a custom JSON Schema that we use in our case), the vscode-quarkus is started after and it cannot register the registerContributor. In other words you need to type something in the YAML editor, wait for compute of MP properties and after that it should work. If you close and reopen the application.yaml you will not have this problem.

 * the classpath changed has been managed. In other words if you modify the classpath (or modify a MP rest client) you should see the new property (please note that value doesn't work at this moment). But you need to change something in the editor because vscode-yaml cannot detect the change of JSON Schema)

src/yaml/YamlSchema.ts
angelozerr: Should I replace with `MicroProfile` ?

fbricon: Just "Quarkus" for now

package.json
angelozerr: Removed, it seems it works without this yaml delclaration.

src/yaml/YamlSchema.ts
angelozerr: updated

package.json
```diff
@@ -36,6 +36,7 @@
     "onCommand:quarkusTools.debugQuarkusProject",
     "onCommand:quarkusTools.welcome",
     "workspaceContains:**/src/main/resources/application.properties",
+    "workspaceContains:**/src/main/resources/application.yaml",
     "onLanguage:quarkus-properties",
     "onLanguage:java",
     "onLanguage:qute"
```
xorye: Should be `"onLanguage:qute-html"`.

If you want to keep `qute`, please change https://github.com/redhat-developer/vscode-quarkus/blob/2bd8d5013ad1bd9f151275af93cd093eb3ad3260/package.json#L66

src/extension.ts
```diff
@@ -37,7 +38,13 @@ export function activate(context: ExtensionContext) {
 
   context.subscriptions.push(createTerminateDebugListener());
 
+  /**
+     * Register Yaml Schema support to manage application.yaml
```
xorye: Please fix indentation

src/extension.ts
```diff
@@ -50,8 +57,10 @@ export function activate(context: ExtensionContext) {
     /**
      * Delegate notifications from Java JDT LS to the MicroProfile LS
      */
-    bindNotification('microprofile/propertiesChanged');
-
+    context.subscriptions.push(commands.registerCommand('microprofile/propertiesChanged', (event: any) => {
```
xorye: IMO, it would be more clear if we did `(event: MicroProfilePropertiesChangeEvent) => {` instead of `(event: any) => {`

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,152 @@
...
+   * @param event the properties change event
+   */
+  public evict(event: MicroProfilePropertiesChangeEvent) {
+    // 1. collect all application.yaml uri which belong to the project URIs
+    const uriToDelete: Array<string> = [];
+    for (const entry of this.cache.entries()) {
+      if (event.projectURIs.includes(entry[1].projectURI)) {
+        uriToDelete.push(entry[0]);
+      }
+    }
```
xorye: I would consider using [`Map.forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach) in this case: 

`this.cache.forEach((value: JsonSchemaForProjectInfo, key: string) => {})`

or maybe

`this.cache.forEach((jsonSchema: JsonSchemaForProjectInfo, uri: string) => {})`

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,152 @@
...
+    // 2. evict the cache
+    uriToDelete.forEach(uri => {
```
xorye: I suggest `uriToDelete.forEach(this.cache.delete);`

package.json
angelozerr: @xorye I did nothing about that, master contains that https://github.com/redhat-developer/vscode-quarkus/blob/master/package.json#L41

I think it should be fixed on an another PR.

xorye:
For this popup:
image

I think it would be nice if there was a button that allowed you to install vscode-yaml without going to the extensions pane. It would be very convenient for the user.

The Java extension pack has similar functionality where you can click a link to automatically install a missing extension:
https://github.com/microsoft/vscode-java-pack/blob/30689fe903d15677ffe0cd45de9d9f15cc604dc6/src/commands/handler.ts#L43-L50

@fbricon @angelozerr Should I create a separate issue for this?

src/extension.ts
angelozerr: fixed

src/extension.ts
angelozerr: fixed

angelozerr:
> @fbricon @angelozerr Should I create a separate issue for this?

Yes please!

xorye:
Please delete `bindNotification()`, https://github.com/redhat-developer/vscode-quarkus/blob/768a26e5da446a83b1d4b0c63262627fc59e7fcc/src/extension.ts#L78-L82

since `bindNotification()` is not being used anymore

package.json
xorye: @angelozerr oops, you are right. I will make a PR for this.

angelozerr: thanks!

src/yaml/YamlSchema.ts
angelozerr: nice, thank's for the suggestion, updated.

src/yaml/YamlSchema.ts
angelozerr: I'm sorry, I don't understand your suggestion?  The idea is to fill uriToDelete, could you explain more your suggestion please. Thanks!

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+    // 2. evict the cache
+    uriToDelete.forEach(this.cache.delete);
+  }
+}
+
+let yamlSchemaCache: YamlSchemaCache;
+
+export async function registerYamlSchemaSupport(): Promise<YamlSchemaCache> {
```
xorye: Please change the return type to `Promise<YamlSchemaCache | undefined>`

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,150 @@
...
+  const yamlPlugin: any = await activateYamlExtension();
+  if (!yamlPlugin || !yamlPlugin.registerContributor) {
+    // activateYamlExtension has already alerted to users for errors.
+    return;
```
xorye: `return undefined;`

src/extension.ts
```diff
@@ -37,7 +38,13 @@ export function activate(context: ExtensionContext) {
 
   context.subscriptions.push(createTerminateDebugListener());
 
+  /**
+   * Register Yaml Schema support to manage application.yaml
+   */
+  const yamlSchemaCache = registerYamlSchemaSupport();
+
   connectToLS(context).then(() => {
+    yamlSchemaCache.then(cache => cache.languageClient = languageClient);
```
xorye: Please check if cache is not undefined first `cache => {if (cache) cache.languageClient = languageClient; }`

src/extension.ts
```diff
@@ -50,8 +57,10 @@ export function activate(context: ExtensionContext) {
     /**
      * Delegate notifications from Java JDT LS to the MicroProfile LS
      */
-    bindNotification('microprofile/propertiesChanged');
-
+    context.subscriptions.push(commands.registerCommand('microprofile/propertiesChanged', (event: MicroProfilePropertiesChangeEvent) => {
+      languageClient.sendNotification('microprofile/propertiesChanged', event);
+      yamlSchemaCache.then(cache => cache.evict(event));
```
xorye: Please check if cache is not undefined first `cache => {if (cache) cache.evict(event)}`

src/yaml/YamlSchema.ts
xorye: Sorry, what I meant was to use `forEach()` instead of `for (const entry of this.cache.entries())`.

Like this:
```
this.cache.forEach((jsonSchema: JsonSchemaForProjectInfo, uri: string) => {
    if (event.projectURIs.includes(jsonSchema.projectURI)) {
        uriToDelete.push(uri);
    }
});
```

angelozerr: Ok thanks for the explanation, done.

src/yaml/YamlSchema.ts
angelozerr: done

src/yaml/YamlSchema.ts
angelozerr: done

src/extension.ts
angelozerr: done

src/extension.ts
angelozerr: done

angelozerr:
> Please delete bindNotification(),

@xorye done

fbricon:
It asks me to install the YAML extension on startup, even before I started to open an application.yaml

angelozerr:
Indeed but is it something that @xorye could improve it once the pr is merged? Or perhaps you want that i remove the dialog?

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,150 @@
...
+      return yamlSchema.jsonSchema;
+    }
+    if (!this._languageClient) {
+      return undefined;
+    }
+    const params = {
+      uri: applicationYamlUri,
+      scopes: [1, 2]
+    };
+    return this._languageClient.sendRequest('microprofile/jsonSchemaForProjectInfo', params)
```
fbricon: use a constant

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,150 @@
...
+        const jsonSchema = result.jsonSchema;
+        this.cache.set(applicationYamlUri, result);
+        return jsonSchema;
+      }, (err) => {
+        console.error("Error while consumming 'microprofile/jsonSchemaForProjectInfo' request: " + err.message);
+      });
+  }
+
+  /**
+   * Evict the cache according to the classpath changed of the given project URIs
```
fbricon, line 78: "Error while consuming "+ constant

src/yaml/YamlSchema.ts
```diff
@@ -0,0 +1,150 @@
...
+    uriToDelete.forEach(this.cache.delete);
+  }
+}
+
+let yamlSchemaCache: YamlSchemaCache;
+
+export async function registerYamlSchemaSupport(): Promise<YamlSchemaCache | undefined> {
+  const yamlPlugin: any = await activateYamlExtension();
+  if (!yamlPlugin || !yamlPlugin.registerContributor) {
+    // activateYamlExtension has already alerted to users for errors.
```
fbricon: alerted users

fbricon (changes requested):

src/yaml/YamlSchema.ts
fbricon: about errors

angelozerr: @fbricon fixed

src/yaml/YamlSchema.ts
angelozerr: @fbricon fixed

src/yaml/YamlSchema.ts
angelozerr: @fbricon fixed

fbricon:
When adding a new quarkus extension, after opening application.yaml, there's an error in the log:
```
ejected promise not handled within 1 second: TypeError: Method Map.prototype.delete called on incompatible receiver undefined
extensionHostProcess.js:832
stack trace: TypeError: Method Map.prototype.delete called on incompatible receiver undefined
	at delete (<anonymous>)
	at Array.forEach (<anonymous>)
	at YamlSchemaCache.evict (/Users/fbricon/Dev/projects/vscode-quarkus/dist/extension.js:83053:21)
	at /Users/fbricon/Dev/projects/vscode-quarkus/dist/extension.js:80758:23
	at o (/Users/fbricon/.vscode-insiders/extensions/vscjava.vscode-java-test-0.22.0/dist/extension.bundle.js:9:136323)
	at /Users/fbricon/.vscode-insiders/extensions/vscjava.vscode-java-test-0.22.0/dist/extension.bundle.js:9:136820
```

then completion stops working

Issue #181, opened 2020-01-02
Update language server to support microprofile-config.properties


jclingan:
Quarkus supports both src/main/resources/application.properties and src/main/resources/META-INF/microprofile-config.properties, the latter being the MicroProfile standard configuration file.

The language server supports application.properties but not microprofile-config.properties. This request is to add support for microprofile-config.properties to get syntax highlighting, code completion, etc.

The primary justification is that developers often like to develop portable code, and this request would help to achieve that goal while remaining productive. In  addition, developers often want to try existing projects on Quarkus, and having code completion of existing microprofile-config.properties would greatly facilitate this use case without having to RTFM. Ex: "Re-hosting my project on  Quarkus nearly works, but how the heck do I change the default http port"? Code-completion really helps with scenarios like this!

angelozerr:
@jclingan thank's for having reported this issue.

If I understand correctly `microprofile-config.properties` should have only microprofile properties (ex : MyClass/mp-rest/url) and application.properties should have microprofile properties and Quarkus properties both.

Is that?

jclingan:
Let me add @kenfinnigan. We would want any kind of property (quarkus, application, microprofile) supported by both property files. However, I want to check with Ken (who can pull in others if required) on having both property files at the same time for the same app.  I wouldn't think we would want that.

maxandersen:
Is this anything more than adding an additional filename the editor and validation extension will react to ?

angelozerr:
> Is this anything more than adding an additional filename the editor and validation extension will react to ?

If the idea is to have in `microprofile-config.properties` the same behavior than `application.properties `I think it can be easy to do. As @maxandersen it's just adding a file name is the package.json of the vscode-quarkus.

The only thing that we must manage too is on JDT LS extension side which load application.properties to get the server port (to manage CodeLens URL (base url)). We will need to load the both files.

maxandersen:
codelens url ? what is that for ? 

What from the properties file does JDT LS need to get a base url ? (I assume here JDT LS is just for java content assist/validation but sounds like there is something else at play here ? )

angelozerr:
> codelens url ? what is that for ?

It's for JAX-RS:

![](https://user-images.githubusercontent.com/1932211/68654940-53750080-052f-11ea-99e4-46188577a681.png)

See PR at https://github.com/redhat-developer/quarkus-ls/pull/140#issuecomment-552788862

You can configure the port in` application.properties` with 

[...]

PR #191, opened 2020-01-14
Update language server to support microprofile-config.properties

angelozerr:
Update language server to support microprofile-config.properties

See #181

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,8 @@
     "onCommand:quarkusTools.debugQuarkusProject",
     "onCommand:quarkusTools.welcome",
     "workspaceContains:**/src/main/resources/application.properties",
+    "workspaceContains:**src/main/resources/META-INF/microprofile-config.properties",
+    "onLanguage:microprofile-properties",
     "onLanguage:quarkus-properties",
     "onLanguage:java",
     "onLanguage:qute-html"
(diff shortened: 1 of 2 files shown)


angelozerr:
This PR provides the same support for microprofile-config.properties than Quarkus application.properties. 

Please don't close the issue #181 because there is some work to do on JDT LS extension to manage properties for Java hover and CodeLens URL

package.json
```diff
@@ -95,6 +107,10 @@
         "language": "java",
         "path": "./snippets/quarkus.json"
       },
+      {
+        "language": "microprofile-properties",
+        "path": "./snippets/properties.json"
```
fbricon: nope, current snippets are quarkus specific.

src/extension.ts
```diff
@@ -125,6 +125,7 @@ function connectToLS(context: ExtensionContext) {
   return requirements.resolveRequirements().then(requirements => {
     const clientOptions: LanguageClientOptions = {
       documentSelector: [
+	    { scheme: 'file', language: 'microprofile-properties' },
```
fbricon: indentation

package.json
angelozerr: @fbricon fixed

src/extension.ts
angelozerr: @fbricon fixed

PR #194, opened 2020-01-16
Add Qute TextMate grammar and language configuration for JSON, YAML and txt files

### LINKED ISSUE #182: Add Qute TextMate grammar and language configuration for HTML files

xorye:
fixes #182 

I will write some of my concerns about this PR tomorrow (mostly about how commenting should work).

For the time being, here are some quick demos:

*.qute.json:
image

*.qute.yaml:
image

*.qute.txt:
image

### DIFF
--- a/language-support/qute/qute-json.tmLanguage.json
+++ b/language-support/qute/qute-json.tmLanguage.json
@@ -1,0 +1,29 @@
+{
+  "name": "Qute JSON",
+  "injections": {
+    "L:support.type.property-name.json, L:string.quoted.double.json": {
+      "patterns": [
+        {
+          "include": "grammar.qute#templates"
+        },
+        {
+          "include": "grammar.qute#comment"
+        }
+      ]
+    },
+    "L:meta.structure.dictionary.json": {
+      "patterns": [
+        {
+          "include": "grammar.qute#comment"
+        }
+      ]
+    }
+  },
+  "patterns": [
+    {
+      "include": "source.json"
+    }
+  ],
+  "scopeName": "source.json.qute",
+  "uuid": "6c100627-887e-4550-82a8-a8b38084bfeb"
+}
--- a/language-support/qute/qute-txt.tmLanguage.json
+++ b/language-support/qute/qute-txt.tmLanguage.json
@@ -1,0 +1,10 @@
+{
+  "name": "Qute Text",
+  "patterns": [
+    {
+      "include": "grammar.qute"
+    }
+  ],
+  "scopeName": "text.qute",
+  "uuid": "753080cc-cfc9-4a1e-8f1b-f5c7d0fd5edc"
+}
--- a/package.json
+++ b/package.json
@@ -38,7 +38,10 @@
     "workspaceContains:**/src/main/resources/application.properties",
     "onLanguage:quarkus-properties",
     "onLanguage:java",
-    "onLanguage:qute-html"
+    "onLanguage:qute-html",
+    "onLanguage:qute-json",
+    "onLanguage:qute-yaml",
+    "onLanguage:qute-txt"
   ],
   "main": "./dist/extension",
   "extensionDependencies": [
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -70,6 +73,36 @@
...
+        ],
+        "extensions": [
+          "qute.yaml"
+        ],
+        "configuration": "./language-support/qute/language-configuration.json"
+      },
+      {
+        "id": "qute-txt",
+        "aliases": [
+          "Qute txt"
```
fbricon: Qute Text

fbricon (changes requested):

fbricon:
@xorye LGTM, except for the small nitpick about `Qute txt`. What's the issue with comments?

xorye:
Hi @maxandersen 

In `*.qute.json` files, do you know if comments are valid? If they are, do you know if the comments should be in this format`{! !}`?

And for `*.qute.yaml` files, should they be commented with `{! !}`?

fbricon:
Given:
```
{
    // regular json shouldn't support that kind of comment
    "foo":"bar"
    "bar": "this is {name} speaking"
    {! qute-style comments should work !}
}
```
This renders as:
Screen Shot 2020-01-20 at 11 06 25 AM

xorye:
@fbricon fixed
image

But for the regular json comment, the default json grammar does in fact highlight those comments:
image

It looks like VSCode offers validation for .json files to notify the user that comments are not valid

language-support/qute/qute-txt.tmLanguage.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Qute txt",
```
fbricon: Qute Text

language-support/qute/qute-json.tmLanguage.json
```diff
@@ -0,0 +1,29 @@
...
+      ]
+    }
+  },
+  "patterns": [
+    {
+      "include": "source.json"
+    }
+  ],
+  "scopeName": "source.json.qute",
+  "uuid": "C011BB07-875D-4DEB-9582-0E1FEC0D1E6E"
```
fbricon: see https://stackoverflow.com/a/59791552/753170

xorye: yes, that was me who asked that :)

fbricon:
qute text is broken:
```
this is {name} speaking
{
    // regular text shouldn't crap out like that
    wat this is {that}?
    {! qute-style comments should work !}
}
```
it doesn't like the brackets
Screen Shot 2020-01-20 at 5 33 00 PM

xorye:
After testing out Qute with the Quarkus Quickstarts sample project: https://github.com/quarkusio/quarkus-quickstarts/tree/master/qute-quickstart
it looks like there cannot be a space or newline when using `{` and `}`, 

ie, `{item.price}` is valid, `{ item.price }`, `{ item.price}` and `{item.price }` are not. Same goes for `{#if` and `{#for`,

I pushed changes for the Qute grammar that takes that into account. 

Demo:
image

xorye:
I found a new problem: 
image

Fixing it now

xorye:
^ That problem is a bit tougher to solve.

It would be helpful if documentation for escaping `{` and `}` was finalized: https://github.com/quarkusio/quarkus/issues/6621

Handlebars suffers from a similar problem:
image

The opening `{{` makes it believe that there will be an expression and a closing `}}` that follows it, but since there is no closing `}}`, the grammar thinks that everything until the end of the file is the expression.

If the user truly wanted to have `{{hello` in their HTML, it would be better for the grammar if `&#123;` or `\{` was used instead of `{`.

fbricon:
@xorye please open a separate ticket for the above issue

PR #196, opened 2020-01-21
Improve vscode-yaml integration

### LINKED ISSUE #195: Improve checking vscode-yaml extension
The YAML support was merged, but it must be improved (@xorye please do that) because if you don't have vscode-yaml, the vscode-quarkus shows a popup (please install vscode-yaml).

The improvement should:

 * do the check vscode-yaml only when a application.yaml or application.yml is opened.
 * improve the install popup like vscode-java.

xorye:
fixes #195 

image

After clicking install:
image

The prompt will only ask the user to install/upgrade if the user is currently on an `application.yaml` or `application.yml` file.


xorye:
Updated to use `vscode.env.appName` instead of hardcoding "VS Code"

PR #198, opened 2020-01-23
Have RESTEasy JAX-RS always selected in the extensions picker

### LINKED ISSUE #141: Improve the QuickPick for choosing extensions
Currently, the QuickPick for choosing extensions looks like the following:
image

It would be nice if the "RESTEasy JAX-RS" entry was always checked because RESTEasy JAX-RS is included in all Quarkus projects by default. Currently, the user can check it, or uncheck it. But regardless of if its checked or not, that extension is always included.
image

Also, like https://code.quarkus.io/, it would be nice if there was a "Preview" label for the preview extensions. This information is provided by the code.quarkus.io  /api/extensions endpoint,

Lastly, having predefined categories for the extensions list would also be a nice touch.

xorye:
Fixes #141 

Here is the extensions picker when using the project generation wizard:
image

image

Here is the extensions picker when using the add extensions wizard:
image

The "Last used" option should not show up if the last used extension was only RESTEasy JAX-RS.

For the add extensions wizard, the  RESTEasy JAX-RS is being omitted from the list because this PR assumes that RESTEasy JAX-RS is already present in the current Quarkus project.

Also, "RESTEasy JAX-RS" is being hardcoded
https://github.com/xorye/vscode-quarkus/blob/e73cdc3d1bd24760efe61e809f4533c2f6525558/src/definitions/QExtension.ts#L38

because the api that we get the extensions list from (https://code.quarkus.io/api/extensions) does not specify that RESTEasy JAX-RS is always included.


xorye:
pickExtensions.ts was deleted in this PR because the file is not being used anywhere. I forgot to delete it from before

PR #204, opened 2020-01-31
Determine if current project is a Quarkus project before adding Quarkus extensions.

### LINKED ISSUE #45: Determine if current project is a Quarkus project before adding Quarkus extensions
The extension currently lets the user try to add Quarkus extensions even if a Maven Quarkus project is not open.

It would be great if we had something like this: 
1. Open a folder that does not contain a Quarkus project
2. Invoke the "Quarkus: Add extensions to current project" command (which starts the vscode-quarkus extension since it is one of the [activation events](https://github.com/redhat-developer/vscode-quarkus/blob/a27e083ca4c4c858048253eda6f2651082a75120/package.json#L30))
3. vscode-quarkus then detects that we are not in a Quarkus project, and displays an error message.

xorye:
Fixes #45 

With this PR: https://github.com/redhat-developer/quarkus-ls/pull/205, the projects in the current workspace are being checked if they are Quarkus projects, before the `Quarkus: Add extensions to current project` command becomes available.


src/quarkusProjectListener.ts
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { commands, workspace } from 'vscode';
+import { checkQuarkusProjectExistsWorkspace } from './utils/workspaceUtils';
+
+export function createQuarkusProjectListener() {
+  return workspace.onDidChangeWorkspaceFolders(async () => {
+    await new Promise((res => setTimeout(res, 100)));
```
xorye: This was done because without it, the results from the `java/projectTypes` isn't accurate.

For example, if I have 3 quarkus projects in my workspace and I remove one, the `java/projectTypes`  command gets called, and it returns information that indicates that I still have 3 quarkus projects in my workspace, when I only have 2.

xorye:
Works on Windows 10

xorye:
The `microprofile/java/projectLabels` command is called whenever the workspace folders are changed (new folder is added or removed from the workspace) and makes the `Quarkus: Add extensions to current project` command availible accordingly.

If there are no Quarkus projects in the workspace, the command appears in the command palette (Ctrl + Shfit + P). If there are Quarkus projects, the command appears.

xorye:
In my latest push, per @angelozerr 's suggestion, I am also retrieving project labels when the `microprofile/propertiesChanged` notification is sent from the MicroProfile ls to vscode-quarkus .

Project labels are only being retrieved when receiving the `microprofile/propertiesChanged` notification IF the `type` indicates a dependency change: https://github.com/redhat-developer/quarkus-ls/blob/0ff91b4d0fe4670584a3ad23fe645c8141df7f3d/microprofile.ls/com.redhat.microprofile.ls/src/main/java/com/redhat/microprofile/commons/MicroProfilePropertiesChangeEvent.java#L23

Therefore as a result, if you change your Quarkus project's pom.xml so that it is not longer a Quarkus project, the `Quarkus: Add extensions to current Quarkus project` command should not appear.

Demo:
Gif showing the command not appearing after changing pom.xml

xorye:
Please do not merge this yet, I am quickly editing it so that we can have: https://github.com/redhat-developer/vscode-quarkus/issues/215 in this PR, but with parent folder name instead of project name.

xorye:
Done:
image

angelozerr:
@xorye everything works very good (with maven and gradle) except for multiple project.

I have the same window than you:

image

But when I try to select a project, it does nothing. I have not tried to debug it. if you have some information how to debug it, please tell me.

xorye:
@angelozerr fixed

angelozerr:
@xorye works like a charm.

PR #212, opened 2020-02-18
Remove onLanguage:java activation event

xorye:

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "workspaceContains:**/src/main/resources/application.yml",
     "onLanguage:microprofile-properties",
     "onLanguage:quarkus-properties",
-    "onLanguage:java",
     "onLanguage:qute-html",
     "onLanguage:qute-json",
     "onLanguage:qute-yaml",


angelozerr:
@xorye could you explain please what you do that. If I remember I did that to manage codelens and hover in java files.

xorye:
The main reason was that if I have a Java file from a regular Java project open (not a Quarkus project), and I try hover, an error message pops out from the 'Quarkus Tools' output:
gif showing the hover error

angelozerr:
Have you the same error once java jdt ls is started? 

IMHO we must discover why we have this bug and try to fix it.

angelozerr:
@xorye I cannot reproduce this issue (screenshot image is not available). Perhaps my refactor with hover as extension point fixes the problem?

xorye:
Here's the gif of the problem I was describing before:
gif showing the hover error

However, now I can't reproduce it either

angelozerr:
> However, now I can't reproduce it either

@xorye I close this issue since problem is gone, but if you have problem again, please reopen it.

PR #217, opened 2020-02-24
Choose Quarkus project to debug when multiple are open

### LINKED ISSUE #74: Cannot select which project to debug when multiple workspace folders are open
Open multiple workspace folders containing Maven Quarkus projects  in VS Code like so:
```
project1/
project2/
project3/
```
Do not open any file (ensure there are no open editors), just have the folders loaded in VS Code.

When calling the debug command (Quarkus: Debug current Quarkus project), we currently do not know which Quarkus project should be debugged.

In this case, I propose to add an input box, similar to what we already have for the Add Extensions wizard:
image
Something like that could help determine which Quarkus project to debug.

xorye:
Fixes #74 

This PR contains two commits.
The first commit implements the feature to choose which Quarkus project to debug:
image

When the debug command is run while there is a file open in VS Code, this PR will check if that file is within a Quarkus project directory. If it is, then that Quarkus project will be debugged. If not, the project chooser dialog (the image above) will appear.

The second commit does some refactoring so that the 'Add extensions' command uses the same project choosing strategy from the first commit, in order to reduce duplicate code. Therefore both 'Debug project' and 'Add extensions' commands will open the same dialog as above.

This PR has been tested on Linux and Windows.

This PR supports two main cases when debugging when there are more than one Quarkus project open in VS Code.

**First case**: When the Quarkus projects are opened in VS Code as separate workspace folders
image
In the image above, a workspace called 'Untitled' is open. It contains two workspace folders, `qm/` and `qm2/`. Both of them each contain a Quarkus project.

In this case, when calling the debugging command, the launch config and tasks will be located in `qm/.vscode/` or `qm2/.vscode/`, depending on which project is being debugged.

**Second case**: When the Quarkus projects are opened in VS Code as a folders inside of a single workspace folder
image
In the image above, a single workspace folder is open (`maven/`), which contains two folders: `qm/` and `qm2/` where both of them each contain a Quarkus project.

In this case, when calling the debugging command, the launch config and tasks for both projects will be located in `maven/.vscode/` folder.

To test this PR, here are some test projects: 
[debug_multiple_test.zip](https://github.com/redhat-developer/vscode-quarkus/files/4246216/debug_multiple_test.zip)


angelozerr:
@xorye I have imported the folder which contains maven and gradle project and when I open the Quarkus debug, I have just gradle projects:

image

Is it a known limitation (perhaps a JDT LS limitation)?

angelozerr:
@xorye it works pretty well, I have tried to add, remove project by using your zip test (thanks for that) and it works well. 

The only issue that I have seen is when you open a project, and you remove it from the workspace, the Quarkus Debug command already exists although it should not exist. 

image

If I select the command, I have an error:

image

xorye:
@angelozerr I am able to reproduce your problem. When the error appears, could you run the `Show Running Extensions` command to see if vscode-quarkus is running? I suspect that when the last workspace folder is removed, vscode-quarkus is being terminated immediately without having the chance to handle the workspace.onDidChangeWorkspaceFolders event.

angelozerr:
> When the error appears, could you run the Show Running Extensions command to see if vscode-quarkus is running?

Indeed it's that. I think we can live with this little bug. @xorye if you find an elegant and easy fix, do it, otherwise don't loose your time about this issue and merge the PR.

xorye:
I have realized that the error can also happen in another case:

1. Open a Quarkus Maven project
2. Wait until vscode-quarkus starts
3. Confirm that the debug command works
4. Add a second Quarkus Maven project to the workspace by from File -> Add Folder to Workspace
    (after adding the folder, I don't know why, but the VSCode window reloads, causing 
     vscode-quarkus to have to activate for a second time)
5. Quickly open the command palette to see that the debug command exists
6. Call the debug command which causes the error

The error happens because this `setContext` command:
https://github.com/xorye/vscode-quarkus/blob/c12e0676105c85fe6bdda7e5ecc06b1b37da148a/src/QuarkusProjectListener.ts#L62
seems to save the `setContext` value across vscode-quarkus restarts.

Thankfully, the  `setContext` value does not persist across VSCode restarts (when VSCode is closed and reopened)

Maybe instead we should have the Debug and Add extensions command to be always available and try to find Quarkus projects in the workspace only after the commands are called?

xorye:
> if you find an elegant and easy fix, do it, otherwise don't loose your time about this issue and merge the PR.

Thank you for the feedback, I am looking into it now

xorye:
Fixed travis failure, merging now

PR #226, opened 2020-03-11
Use staging URL for project generation tests

xorye:
The Travis build now sets a new environment variable: `VSCODE_QUARKUS_API_URL=https://stage.code.quarkus.io/api` for vscode-quarkus to use for the project generation tests.


.travis.yml
```diff
@@ -17,6 +17,9 @@ branches:
   only:
     - master
 
+env:
```
fbricon: The env property should be set by the test-ui-run command instead, so it works from CLI everywhere.

fbricon (changes requested):

src/test/vscodeUiTest/suite/projectGenerationTest.ts
```diff
@@ -31,13 +31,19 @@ use(require('chai-fs'));
  */
 describe('Project generation tests', function() {
```
fbricon, line 32: maybe try process.env.VSCODE_QUARKUS_API_URL=https://stage.code.quarkus.io/api directly here. 
Dunno if it'll work

xorye: fixed

.travis.yml
xorye: fixed

PR #230, opened 2020-03-16
Use source.ini for the properties TextMate grammar

### LINKED ISSUE #229: Update application.properties TextMate grammar for multi-line properties values
For multiline properties like this one:
image

the TextMate grammar highlighting is incorrect.

xorye:
Fixes #229 

Instead of maintaining our own textmate grammar, it would be better if we use VS Code's properties textmate grammar instead: https://github.com/microsoft/vscode/blob/master/extensions/ini/syntaxes/ini.tmLanguage.json

Here's how it looks:
image


fbricon:
Excellent, thanks @xorye! The less code to maintain the better!

PR #235, opened 2020-03-25
Edit TextMate grammar to allow whitespaces before braces

### LINKED ISSUE #221: [Qute] Syntax coloration problem with quote
Take the following Qute template:

```
{#if (condition)}
 'aaaaaaaa'
{/if}

{#if (condition) }
 'aaaaaaaa'
{/if}

{#if (condition) }
 'aaaaaaaa
{/if}

```

Here a screenshot which shows in action quote problem (see the last if):

image

It seems when there is no space between ')' and '}' the `aaaaaaaa` which is a text is not be considered as string (it's OK) but when there is a a space `aaaaaaaa`  is considered as string which is wrong.

xorye:
Fixes #221 

The syntax highlighting does not break when there are whitespaces before the closing brace
image

### DIFF
--- a/language-support/qute/qute.tmLanguage.json
+++ b/language-support/qute/qute.tmLanguage.json
@@ -42,7 +42,7 @@
     },
     "section_start_tag": {
       "begin": "({)(#)(else\\sif|\\w+(\\.\\w+)*)",
-      "end": "(\\/)?((?<![\\s\\\\])})",
+      "end": "(\\/)?((?<![\\\\])})",
       "beginCaptures": {
         "1": {
           "name": "support.constant.handlebars"
@@ -70,7 +70,7 @@
     },
     "section_end_tag": {
       "begin": "({)(\\/)(\\w+(\\.\\w+)*)?\\s*",
-      "end": "((?<![\\s\\\\])})",
+      "end": "((?<![\\\\])})",
       "beginCaptures": {
         "1": {
           "name": "support.constant.handlebars"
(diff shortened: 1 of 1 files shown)


fbricon:
Ok it works, but I see that for instance `{name }` breaks too. I'd really like to see what the proper rule for handling whitespaces next to brackets is, in qute. cc @mkouba

xorye:
Yes, this PR only takes care of the case I show above. It's very easy to fix all cases where whitespaces exist before/after the Qute braces. Although, as I mentioned in https://github.com/quarkusio/quarkus/issues/8158, I'm not sure if its correct to have spacing in those situations.

Should I do the TextMate grammar fixes in this PR so that syntax highlighting doesn't break?

fbricon:
let me know when the grammar is fixed to allow whitespaces for all relevant cases

xorye:
In my latest push, the grammar now allows whitespace before closing braces:
image

```
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<title>Report {now}</title>
</head>
<body>
    <h1>Report {now   }</h1>
    {#for sample in test.fname()     }
      <h2>{sample.name ?: 'Unkn{hello  }own'         }</h2>
      <p>
      {#if sample.valid > 1100      }
        {sample.data}
        {myObject.myValue}
      {#else     }
        <strong>Invalid sample found</strong>.
      {/if     }
      </p>
    {/for       }
    First name:
    <input type="text" name="fname" class="{test    }"
      value="{myObject.myValue       }"/>
    <br>
    <h1>Hi: {myObject.myValue        } </h1>
    </div>
</body>
</html>
```

mkouba:
So although it's supported it is not recommended and maybe we should encourage the users to remove the whitespace? I'm not sure about the possibilities we have (tooltip, warning, etc.).

BTW it looks really nice!

fbricon:
@mkouba that's the job for a language server to provide linting for qute syntax. See https://github.com/redhat-developer/quarkus-ls/issues/176

fbricon:
Thanks @xorye !

PR #244, opened 2020-04-15
Fix tslint trailing whitespace problem

angelozerr:
Fix tslint trailing whitespace problem


fbricon:
Thanks @angelozerr!

PR #252, opened 2020-04-30
Fix terminal not appearing when adding extensions with the Add Extensions command

xorye:


fbricon:
LGTM but please also update the changelog

xorye:
Updated the changelog

fbricon:
Can't reproduce the issue with the fix, nice work @xorye!

PR #253, opened 2020-05-04
Include microprofile.ls and quarkus.ls.ext jar

rzgry:
Include the `quarkus.ls.ext` jar added in https://github.com/redhat-developer/quarkus-ls/pull/304

Part of https://github.com/redhat-developer/quarkus-ls/issues/295

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,14 +15,24 @@ const gulp = require('gulp');
 const rename = require('gulp-rename');
 const cp = require('child_process');
 
-const serverName = 'com.redhat.microprofile.ls-uber.jar';
+const microprofileServerName = 'com.redhat.microprofile.ls-uber.jar';
 const extensions = ['com.redhat.microprofile.jdt.core', 'com.redhat.microprofile.jdt.quarkus'];
-const serverDir = '../quarkus-ls/microprofile.ls/com.redhat.microprofile.ls';
+const microprofileServerDir = '../quarkus-ls/microprofile.ls/com.redhat.microprofile.ls';
 const extensionDir = '../quarkus-ls/microprofile.jdt';
 
-gulp.task('buildServer', (done) => {
-  cp.execSync(mvnw() + ' clean verify -DskipTests', { cwd: serverDir , stdio: 'inherit' });
-  gulp.src(serverDir + '/target/' + serverName)
+const quarkusServerExtName = 'com.redhat.quarkus.ls-0.0.1-SNAPSHOT.jar';
+const quarkusServerExtDir = '../quarkus-ls/quarkus.ls.ext/com.redhat.quarkus.ls'
+
+gulp.task('buildMicroProfileServer', (done) => {
+  cp.execSync(mvnw() + ' clean verify -DskipTests', { cwd: microprofileServerDir , stdio: 'inherit' });
+  gulp.src(microprofileServerDir + '/target/' + microprofileServerName)
+    .pipe(gulp.dest('./server'));
+  done();
+});
+
+gulp.task('buildQuarkusServerExt', (done) => {
+  cp.execSync(mvnw() + ' clean verify -DskipTests', { cwd: quarkusServerExtDir , stdio: 'inherit' });
+  gulp.src(quarkusServerExtDir + '/target/' + quarkusServerExtName)
     .pipe(gulp.dest('./server'));
   done();
 });
@@ -37,7 +47,7 @@ gulp.task('buildExtension', (done) => {
   done();
 });
 
-gulp.task('build', gulp.series('buildServer', 'buildExtension'));
+gulp.task('build', gulp.series('buildMicroProfileServer', 'buildQuarkusServerExt', 'buildExtension'));
 
 function mvnw() {
 	return isWin() ? 'mvnw.cmd' : './mvnw';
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -15,14 +15,24 @@ const gulp = require('gulp');
...
+const microprofileServerName = 'com.redhat.microprofile.ls-uber.jar';
 const extensions = ['com.redhat.microprofile.jdt.core', 'com.redhat.microprofile.jdt.quarkus'];
-const serverDir = '../quarkus-ls/microprofile.ls/com.redhat.microprofile.ls';
+const microprofileServerDir = '../quarkus-ls/microprofile.ls/com.redhat.microprofile.ls';
 const extensionDir = '../quarkus-ls/microprofile.jdt';
 
-gulp.task('buildServer', (done) => {
-  cp.execSync(mvnw() + ' clean verify -DskipTests', { cwd: serverDir , stdio: 'inherit' });
-  gulp.src(serverDir + '/target/' + serverName)
+const quarkusServerExtName = 'com.redhat.quarkus.ls-0.0.1-SNAPSHOT.jar';
```
fbricon: You can't hard code the version, better use glob patterns to catch the jar, regardless of its version

src/languageServer/javaServerStarter.ts
```diff
@@ -37,9 +39,12 @@ function prepareParams(): string[] {
...
-  const launchersFound: Array<string> = glob.sync(`**/${SERVER_NAME}`, { cwd: serverHome });
-  if (launchersFound.length) {
-    params.push('-jar'); params.push(path.resolve(serverHome, launchersFound[0]));
+  const microprofileServerFound: Array<string> = glob.sync(`**/${MICROPROFILE_SERVER_NAME}`, { cwd: serverHome });
+  const quarkusServerExtFound: Array<string> = glob.sync(`**/${QUARKUS_SERVER_EXTENSION_NAME}`, { cwd: serverHome });
+  if (microprofileServerFound.length && quarkusServerExtFound.length) {
+    params.push('-cp');
+    params.push(`${serverHome}/*`);
+    params.push(MICROPROFILE_SERVER_MAIN_CLASS);
   }
```
fbricon, line 50: please log something in case no jars were found

src/languageServer/javaServerStarter.ts
```diff
@@ -7,7 +7,9 @@ const glob = require('glob');
 
 const DEBUG = startedInDebugMode();
 const DEBUG_PORT = 1064;
-const SERVER_NAME = 'com.redhat.microprofile.ls-uber.jar';
+const MICROPROFILE_SERVER_NAME = 'com.redhat.microprofile.ls-uber.jar';
+const QUARKUS_SERVER_EXTENSION_NAME = 'com.redhat.quarkus.ls-0.0.1-SNAPSHOT.jar';
```
fbricon: don't include the version

gulpfile.js
```diff
@@ -20,7 +20,7 @@ const extensions = ['com.redhat.microprofile.jdt.core', 'com.redhat.microprofile
 const microprofileServerDir = '../quarkus-ls/microprofile.ls/com.redhat.microprofile.ls';
 const extensionDir = '../quarkus-ls/microprofile.jdt';
 
-const quarkusServerExtName = 'com.redhat.quarkus.ls-0.0.1-SNAPSHOT.jar';
+const quarkusServerExtGlob = 'com.redhat.quarkus.ls!(*-sources).jar';
```
angelozerr, line 23: @fbricon I see that you have spoken about glob, but why can we not generate Quarkus JAR without a version like the current uber JAR? It will be consistent with the uber JAR of MicroProfile, no?

fbricon: because it's not necessary. we can just drop any jar in there and it'll be picked up, if done properly

fbricon: we could also add the version in the uber jar now, version-less was just more convenient when we had only 1 jar to deal with.

src/languageServer/javaServerStarter.ts
```diff
@@ -37,9 +39,14 @@ function prepareParams(): string[] {
...
-  if (launchersFound.length) {
-    params.push('-jar'); params.push(path.resolve(serverHome, launchersFound[0]));
+  const microprofileServerFound: Array<string> = glob.sync(`**/${MICROPROFILE_SERVER_NAME}`, { cwd: serverHome });
+  const quarkusServerExtFound: Array<string> = glob.sync(`**/${QUARKUS_SERVER_EXTENSION_GLOB}`, { cwd: serverHome });
+  if (microprofileServerFound.length && quarkusServerExtFound.length) {
+    params.push('-cp');
+    params.push(`${serverHome}/*`);
+    params.push(MICROPROFILE_SERVER_MAIN_CLASS);
+  } else {
+    console.error('Unable to find required Language Server JARs');
```
fbricon: we should throw an exception here, because that's a rather serious issue ;-)

gulpfile.js
```diff
@@ -15,18 +15,34 @@ const gulp = require('gulp');
...
 const extensionDir = '../quarkus-ls/microprofile.jdt';
 
-gulp.task('buildServer', (done) => {
-  cp.execSync(mvnw() + ' clean verify -DskipTests', { cwd: serverDir , stdio: 'inherit' });
-  gulp.src(serverDir + '/target/' + serverName)
+const quarkusServerExtGlob = 'com.redhat.quarkus.ls!(*-sources).jar';
+const quarkusServerExtDir = '../quarkus-ls/quarkus.ls.ext/com.redhat.quarkus.ls'
+
+gulp.task('buildMicroProfileServer', (done) => {
+  cp.execSync(mvnw() + ' clean verify -DskipTests', { cwd: microprofileServerDir , stdio: 'inherit' });
```
fbricon: clean install -DskipTests

PR #270, opened 2020-06-29
Display project name when adding extensions

### LINKED ISSUE #215: Display project name when adding extensions
If there are two Quarkus projects in the workspace: quarkus-project-maven and quarkus-project-maven2, the quickpick after calling `Quarkus: Add extension to current Quarkus Project`, this quickpick appears:
image

This enhancement idea adds the project name alonsgide the path to pom.xml/build.gradle:
image

To get the project name, maybe we could have the `microprofile/java/projectLabels` command return not only the project uri, but the project name as well?

datho7561:
When picking which project to add more extensions to when
you have multiple projects in the same workspace,
the name of the project is now displayed,
instead of just the name of the folder that the project is in.

Fixes #215

### DIFF
--- a/src/definitions/ProjectLabelInfo.ts
+++ b/src/definitions/ProjectLabelInfo.ts
@@ -21,10 +21,12 @@ import { WORKSPACE_LABELS_COMMAND_ID } from "./constants";
 
 export class ProjectLabelInfo {
   uri: string;
+  name: string;
   labels: ProjectLabel[];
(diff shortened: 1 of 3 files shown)


datho7561:
Still needs tests

datho7561:
Relies on https://github.com/redhat-developer/quarkus-ls/pull/330 being merged

xorye (approved):

PR #269, opened 2020-06-26
Upgrade vscode-extension-tester to 3.0.2

xorye:
Fixes #266 

Signed-off-by: David Kwon <dakwon@redhat.com>

PR #271, opened 2020-06-29
Set extensions directory for UI tests

### LINKED ISSUE #267: Provide an extension directory when running UI tests
Since vscode-extension-tester version 2.5: https://github.com/redhat-developer/vscode-extension-tester/blob/master/CHANGELOG.md#250-march-20, it was made possible to explicitly provide a directory where extensions are to be installed for the test instance of VS Code.

Currently, the test instance of VS Code detects and installs extensions in the `~/.vscode/extensions` directory, which is the default extensions directory for VS Code.

It would be better to install the test instance's extensions somewhere else, so that it doesn't conflict with already-installed extensions located in `~/.vscode/extensions`.

xorye:
Fixes #267

https://github.com/redhat-developer/vscode-quarkus/pull/269 should be merged before this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -290,8 +290,7 @@ To debug the UI tests, open the VS Code Debugging tab and select
 the "VS Code UI Extension Tests (vscode-quarkus)" at the top left:
 ![](images/runDebuggerUiTests.png)  
 
-When running the UI tests, it would be best to disable all other VS Code extensions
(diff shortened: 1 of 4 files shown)


datho7561 (approved):
This looks good and works on my machine.

fbricon:
conflicts need to be resolved

xorye:
Fixed conflict

datho7561 (approved):

PR #273, opened 2020-07-02
Automatically re-assign the proper language to detected Quarkus properties

### LINKED ISSUE #254: Mitigate vscode-quarkus & springboot tools competing for application.properties
application.properties (and .yaml) are used both by Quarkus and SpringBoot applications for declaring configuration. 
Both vscode-quarkus and springboot-tools extensions use similar file patterns to define the file's language:

https://github.com/redhat-developer/vscode-quarkus/blob/daadbe75acd01cdc82227215669c2301a51aea7a/package.json#L69-L76

vs 
[spring-boot/package.json](https://github.com/spring-projects/sts4/blob/5f25259530a39f2bb1b6dc01c154e45d592c0743/vscode-extensions/vscode-spring-boot/package.json#L41-L66):
```json
    "languages": [
      {
        "id": "spring-boot-properties-yaml",
        "aliases": [
          "Spring Boot Properties Yaml"
        ],
        "filenamePatterns": [
          "application*.yml",
          "application*.yaml",
          "bootstrap*.yml",
          "bootstrap*.yaml"
        ],
        "configuration": "./yaml-support/language-configuration.json"
      },
      {
        "id": "spring-boot-properties",
        "aliases": [
          "Spring Boot Properties"
        ],
        "filenamePatterns": [
          "application*.properties",
          "bootstrap*.properties"
        ],
        "configuration": "./properties-support/language-configuration.json"
      }
    ],
```

When both extensions are installed and enabled, spring's application.properties might be detected as Quarkus properties, and vice-versa, it's unclear which extension takes precedence (I've seen both cases).

It's pretty easy, but not necessarily obvious, that one can switch to the proper file type/language via the vscode status bar. 

So it'd be helpful if there was a way to accurately detect the type of properties we're dealing with, to prevent both extensions from overlapping.

Detecting the project dependencies would require server-side computing, but AFAIK, the only way to do the language distinction is via declarative setting in package.json.

If each framework provided identifiable elements on the 1st line of their application properties, we might be able to make that distinction by leveraging the firstLine property, like in [xml](https://github.com/microsoft/vscode/blob/master/extensions/xml/package.json#L74):
> "firstLine" : "(\\<\\?xml.*)|(\\<svg)|(\\<\\!doctype\\s+svg)"

So having `# quarkus` vs `# springboot` on the 1st line *might* help. But I'm afraid the generic glob patterns will take precedence 1st, so we'll be back to square 1. And the existing thousands of projects out there will not have that identifier anyway.

application.yaml are not defined as quarkus files, since we're simply contributing a custom schema to vscode-yaml, but springboot tools will definitely take over.

We might just need vscode to provide more facilities to identify/register languages.

So there, let's discuss @martinlippert, @aeschli, @emmanuelbernard, @philwebb

angelozerr:
Automatically re-assign the proper language to detected Quarkus properties

Fixes #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The following commands are supported for both Maven and Gradle Quarkus projects:
 
   * `Quarkus: Generate a Quarkus project`: Generate a Quarkus project, based on https://code.quarkus.io/
   * `Quarkus: Add extensions to current project`: Add Quarkus extensions to currently opened Quarkus project
-  * `Quarkus: Debug current Quarkus project`: Launches the Maven `quarkus:dev` plugin or the Gradle `quarkusDev` command and automatically attaches a debugger  
+  * `Quarkus: Debug current Quarkus project`: Launches the Maven `quarkus:dev` plugin or the Gradle `quarkusDev` command and automatically attaches a debugger
 
 ## Quarkus/MicroProfile `properties` Features
 
@@ -34,14 +34,14 @@ In `application.properties` and `microprofile-config.properties` files, you will
   * Code snippets:
     * **qds** - Configure a Quarkus datasource
     * **qj** - Configure a Jaeger tracer
-  
+
 ## Quarkus `yaml` Features
 
 In `application.yaml` and `application.yml` files, you will benefit with:
 
   * Completion support for Quarkus/MicroProfile properties
   * Hover support for Quarkus/MicroProfile properties
-  * Validation support for Quarkus/MicroProfile properties 
+  * Validation support for Quarkus/MicroProfile properties
   * Support for Quarkus profiles
   * Outline support
 
@@ -69,7 +69,7 @@ In `Java` files, you will benefit with:
 ## Supported VS Code settings
 
 The following settings are supported:
-  
+
 * `quarkus.tools.alwaysShowWelcomePage` : Determines whether to show the welcome page on extension startup.
 * `quarkus.tools.debug.terminateProcessOnExit` : Determines whether to terminate the quarkus:dev task after closing the debug session.
 * `quarkus.tools.formatting.surroundEqualsWithSpaces` : Insert spaces around the equals sign when formatting the application.properties file. Default is `false`.
@@ -85,16 +85,16 @@ Default is `none`.
 * `quarkus.tools.validation.unknown.severity` : Validation severity for unknown properties for Quarkus/MicroProfile `*.properties` files. Default is `warning`.
 * `quarkus.tools.validation.unknown.excluded` : Array of properties to ignore for unknown Quarkus properties validation. Patterns can be used ('*' = any string, '?' = any character).
 Default is `["*/mp-rest/providers/*/priority", "mp.openapi.schema.*"]`.
-
-
-Since 1.3.0:
 * `quarkus.tools.codeLens.urlCodeLensEnabled` : Enable/disable the URL code lenses for REST services. Default is`true`.
 * `quarkus.tools.starter.showExtensionDescriptions`: Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions. Default is `true`.
 * `quarkus.tools.validation.value.severity`: Validation severity for property values for Quarkus/MicroProfile `*.properties` files. Default is `error`.
 
+Since 1.6.0:
+* `quarkus.tools.propertiesLanguageMismatch` : Action performed when detected Quarkus properties have an incorrect language. Default is `forceQuarkus`.
+
 ### **Note for MicroProfile Rest Client properties**:
 
-Due to [this issue](https://github.com/redhat-developer/quarkus-ls/issues/203), the MP Rest property: `<mp-rest-client-class>/mp-rest/providers/<mp-rest-provider-class>/priority` reports an unknown error. 
+Due to [this issue](https://github.com/redhat-developer/quarkus-ls/issues/203), the MP Rest property: `<mp-rest-client-class>/mp-rest/providers/<mp-rest-provider-class>/priority` reports an unknown error.
 
 To avoid having this error, you must configure the following in `settings.json`:
 
@@ -124,8 +124,8 @@ CI builds can be installed manually by following these instructions:
 
   1) Download the latest development VSIX archive [from here](https://download.jboss.org/jbosstools/vscode/snapshots/vscode-quarkus/?C=M;O=D). `(vscode-quarkus-XXX.vsix)`
 
-  2) Click `View/Command Palette` 
-  
+  2) Click `View/Command Palette`
+
   3) Type 'VSIX'
 
   4) Select 'Install from VSIX...' and choose the `.vsix` file.
--- a/package.json
+++ b/package.json
@@ -178,6 +178,22 @@
           "description": "Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions.",
           "scope": "window"
         },
+        "quarkus.tools.propertiesLanguageMismatch": {
+          "type": "string",
+          "enum": [
+            "ignore",
+            "forceQuarkus",
+            "prompt"
+          ],
+          "enumDescriptions": [
+            "ignore language mismatch for Quarkus properties",
+            "Automatically correct the language for mismatched Quarkus properties",
+            "Prompt correcting the language for mismatched Quarkus properties"
+          ],
+          "default": "forceQuarkus",
+          "description": "Action performed when detected Quarkus properties have an incorrect language.",
+          "scope": "window"
+        },
         "quarkus.tools.formatting.surroundEqualsWithSpaces": {
           "type": "boolean",
           "default": false,
--- a/src/definitions/ProjectLabelInfo.ts
+++ b/src/definitions/ProjectLabelInfo.ts
@@ -17,7 +17,7 @@ import { commands } from "vscode";
 import { BuildSupport } from "../buildSupport/BuildSupport";
 import { GradleBuildSupport } from "../buildSupport/GradleBuildSupport";
 import { MavenBuildSupport } from "../buildSupport/MavenBuildSupport";
-import { WORKSPACE_LABELS_COMMAND_ID } from "./constants";
+import { JavaVSCodeCommands } from "./constants";
 
 export class ProjectLabelInfo {
   uri: string;
@@ -29,7 +29,7 @@ export class ProjectLabelInfo {
     this.name = name;
   }
 
-  public getBuildSupport(): BuildSupport|undefined {
+  public getBuildSupport(): BuildSupport | undefined {
     if (this.isGradleProject()) {
       return new GradleBuildSupport();
     }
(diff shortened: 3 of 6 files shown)


angelozerr:
This PR force the language ID to `quarkus-properties` for `application.properties` file which belongs to a Quarkus project. If you try for instance to change the language with vscode UI to an another language,  it will not work (this PR reforce the manguage to  `quarkus-properties` )

@fbricon I wonder if we should customize this behavior with settings like:

 * ignore (never forced the languaged ID to `quarkus-properties` )
 * force
 * prompt : before forcing, we have a prompt.

src/extension.ts
```diff
@@ -47,6 +48,35 @@ export function activate(context: ExtensionContext) {
...
   );
 
+  // When extension is started, loop for each text documents which are opened to update their language ID.
+  workspace.textDocuments.forEach(document => {
+    updateLanguageId(document);
+  });
+  // When a text document is opened,  update their language ID.
+  workspace.onDidOpenTextDocument((document) => {
+    updateLanguageId(document);
+  });
```
xorye: Since `workspace.onDidOpenTextDocument` subscribes to an event, we must put it in `context.subsciptions` so that it is disposed when the extension is deactivated

```
context.subscriptions.push(
    workspace.onDidOpenTextDocument((document) => {
        updateLanguageId(document);
  });
);
```

src/definitions/ProjectLabelInfo.ts
```diff
@@ -54,11 +54,22 @@ export class ProjectLabelInfo {
...
     });
   }
+
+  /**
+   * Returns the label information for the given file URI.
+   * @param uri the file URI
+   * @return the label information for the given file URI.
+   */
+  public static async getProjectLabelInfo(uri: string): Promise<ProjectLabelInfo> {
+    const params = { uri: uri };
```
xorye: `const params = { uri };`

src/extension.ts
angelozerr: thanks for your explanation fixed.

src/definitions/ProjectLabelInfo.ts
angelozerr: fixed

angelozerr:
This PR takes care of YAML file too, please review it.

I think settings name and show message should be improved.

package.json
```diff
@@ -178,6 +178,17 @@
           "description": "Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions.",
           "scope": "window"
         },
+        "quarkus.tools.forceToQuarkusLanguage": {
```
fbricon: "quarkus.tools.propertiesLanguageMismatch"

package.json
```diff
@@ -178,6 +178,17 @@
...
+          "type": "string",
+          "enum": [
+            "never",
+            "always",
```
fbricon: "forceQuarkus"

package.json
```diff
@@ -178,6 +178,17 @@
...
+            "prompt"
+          ],
```
fbricon, line 187: "enumDescriptions": [
            "ignore language mismatch for Quarkus properties",
            "Automatically correct the language for mismatched Quarkus properties",
            "Prompt correcting the language for mismatched Quarkus properties"
          ]

package.json
```diff
@@ -178,6 +178,17 @@
...
+          "default": "always",
+          "description": "Force to Quarkus language.",
```
fbricon: Action performed when detected Quarkus properties have an incorrect language

package.json
```diff
@@ -178,6 +178,17 @@
           "description": "Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions.",
           "scope": "window"
         },
+        "quarkus.tools.forceToQuarkusLanguage": {
+          "type": "string",
+          "enum": [
+            "never",
```
fbricon: ignore

fbricon (changes requested):

package.json
angelozerr: fixed

package.json
angelozerr: fixed

angelozerr: fixed

package.json
angelozerr: fixed

package.json
angelozerr: fixed

src/extension.ts
```diff
@@ -47,6 +48,67 @@ export function activate(context: ExtensionContext) {
...
+   * @param languageId the language ID.
+   */
+  async function tryToForceLanguageId(document: TextDocument, languageId: string) {
+    const propertiesLanguageMismatch: PropertiesLanguageMismatch = QuarkusConfig.getPropertiesLanguageMismatch();
+    switch (propertiesLanguageMismatch) {
+      case PropertiesLanguageMismatch.ignore:
+        return;
+      case PropertiesLanguageMismatch.prompt:
+        const YES = "Yes", NO = "No";
+        const response: string | undefined = await window.showInformationMessage(`Do you want to force the language ID of the file  to '${languageId}' to benefit with Quarkus support?`, YES, NO);
```
angelozerr: @fbricon @xorye have you some suggestion how to improve this message ?

datho7561: How about: _Do you want to recognize this file as `${languageId}` and prevent it from being recognized as a different file type? This will improve the Quarkus support._

xorye: My idea is:
```
The language ID for ${filename} must be ${languageId} to receive language support such as
completion and validation. Set the language ID to ${languageId}?
```

image

README.md
```diff
@@ -86,15 +86,17 @@ Default is `none`.
...
 Default is `["*/mp-rest/providers/*/priority", "mp.openapi.schema.*"]`.
 
-
 Since 1.3.0:
 * `quarkus.tools.codeLens.urlCodeLensEnabled` : Enable/disable the URL code lenses for REST services. Default is`true`.
 * `quarkus.tools.starter.showExtensionDescriptions`: Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions. Default is `true`.
 * `quarkus.tools.validation.value.severity`: Validation severity for property values for Quarkus/MicroProfile `*.properties` files. Default is `error`.
 
+Since 1.4.0:
+* `quarkus.tools.propertiesLanguageMismatch` : Action performed when detected Quarkus properties have an incorrect language. Default is`forceQuarkus`.
```
xorye: Default is `forceQuarkus`
(There is a space missing in the current version)

src/extension.ts
angelozerr: Thanks @xorye for your suggestion. Fixed.

README.md
angelozerr: fixed

src/extension.ts
```diff
@@ -47,6 +49,69 @@ export function activate(context: ExtensionContext) {
...
+    }
+  }
+
+  /**
+   * Try to force the document language ID to the given language ID.
+   *
+   * @param document the text document.
+   * @param languageId the language ID.
+   */
+  async function tryToForceLanguageId(document: TextDocument, fileName: string, languageId: string) {
```
fbricon: if not a quarkus project, we should bail here

src/extension.ts
```diff
@@ -47,6 +49,69 @@ export function activate(context: ExtensionContext) {
     })
   );
 
+  // When extension is started, loop for each text documents which are opened to update their language ID.
+  workspace.textDocuments.forEach(document => {
+    updateLanguageId(document);
+  });
+  // When a text document is opened,  update their language ID.
+  context.subscriptions.push(
+    workspace.onDidOpenTextDocument((document) => {
```
fbricon, line 58: check for quarkus project here

src/extension.ts
angelozerr: I think we should keep like I have done, because if we must check if it's a quarkus project, it means `ProjectLabelInfo.getProjectLabelInfo` whoich can takes some times must be called

src/extension.ts
angelozerr: As I add a comment after, checking quarkus project can take some times, so we must do that at then end.

fbricon: right, we should make sure we're dealing with potential problematic files first

src/extension.ts
fbricon: If preference is set to `ignore`, then no need to do the check, just bail.
Else, we need to perform the quarkus check *before* we ask the user. Else, if you open a spring-boot project, you'll be asked to switch the file's language, then check if it's quarkus and do nothing, which makes no sense

angelozerr: > If preference is set to ignore, then no need to do the check, just bail.

fixed

> if you open a spring-boot project, you'll be asked to switch the file's language, then check if it's quarkus and do nothing, which makes no sense

Good catch, fixed

src/extension.ts
```diff
@@ -47,6 +49,72 @@ export function activate(context: ExtensionContext) {
...
+   * @param languageId the language ID.
+   */
+  async function tryToForceLanguageId(document: TextDocument, fileName: string, propertiesLanguageMismatch: PropertiesLanguageMismatch, languageId: string) {
+    // Get project label information for the given file URI
+    const labelInfo = ProjectLabelInfo.getProjectLabelInfo(document.uri.toString());
+    return labelInfo.then(l => {
+      if (l.isQuarkusProject()) {
+        if (propertiesLanguageMismatch === PropertiesLanguageMismatch.prompt) {
+          const YES = "Yes", NO = "No";
+          const response: Thenable<string> = window.showInformationMessage(`The language ID for '${fileName}' must be '${languageId}' to receive language support such as
```
fbricon: to receive Quarkus support

angelozerr: fixed

fbricon:
please fix the conflict

README.md
```diff
@@ -86,15 +86,17 @@ Default is `none`.
 * `quarkus.tools.validation.unknown.excluded` : Array of properties to ignore for unknown Quarkus properties validation. Patterns can be used ('*' = any string, '?' = any character).
 Default is `["*/mp-rest/providers/*/priority", "mp.openapi.schema.*"]`.
 
-
 Since 1.3.0:
 * `quarkus.tools.codeLens.urlCodeLensEnabled` : Enable/disable the URL code lenses for REST services. Default is`true`.
 * `quarkus.tools.starter.showExtensionDescriptions`: Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions. Default is `true`.
 * `quarkus.tools.validation.value.severity`: Validation severity for property values for Quarkus/MicroProfile `*.properties` files. Default is `error`.
 
+Since 1.4.0:
```
fbricon: Since 1.6.0

README.md
```diff
@@ -86,15 +86,17 @@ Default is `none`.
 * `quarkus.tools.validation.unknown.excluded` : Array of properties to ignore for unknown Quarkus properties validation. Patterns can be used ('*' = any string, '?' = any character).
 Default is `["*/mp-rest/providers/*/priority", "mp.openapi.schema.*"]`.
 
-
 Since 1.3.0:
```
fbricon: you can remove "Since 1.3.0"

angelozerr: fixed

README.md
angelozerr: fixed

angelozerr:
> please fix the conflict

done

PR #276, opened 2020-07-06
Build quarkus jdt extension

rzgry:
Builds the quarkus jdt extension that will be added to quarkus-ls in https://github.com/redhat-developer/quarkus-ls/pull/331

Also I updated the glob pattern when building the jdt extensions from `-*.jar` to `-!(*sources).jar` to explicitly ignore the source jar. Since I noticed both the regular jar and sources jar were being picked up by the pattern, but both jars are renamed to the same name so one gets overwritten.


fbricon:
Thanks @rzgry!

PR #295, opened 2020-07-20
Notification when languageID is forced

### LINKED ISSUE #285: Notification when the `quarkus.tools.propertiesLanguageMismatch` setting is set to `forceQuarkus`
When `quarkus.tools.propertiesLanguageMismatch = forceQuarkus`, it would be great if there is a notification to the user if the user tries to set the languageId of an application.properties file from `Quarkus Properties` to something else.

Something along the lines of:
```
Quarkus Tools for Visual Studio Code requires this file's language to be 'Quarkus Properties'
for language support. Edit settings to configure.

[Configure]
```
with a button (`[Configure]`) that takes the user to `quarkus.tools.propertiesLanguageMismatch` in the settings UI.

datho7561:
Provides a notification when the user tries to switch the language ID of the properties file, and provides a 'Configure' button that navigates to the `quarkus.tools.propertiesLanguageMismatch` setting.

Closes #285

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,7 +15,7 @@
  */
 import * as path from 'path';
 import { VSCodeCommands } from './definitions/constants';
-import { ExtensionContext, commands, window, workspace, Terminal, languages, TextDocument } from 'vscode';
+import { ExtensionContext, commands, window, workspace, Terminal, languages, TextDocument, ConfigurationChangeEvent } from 'vscode';
 import { QuarkusContext } from './QuarkusContext';
 import { addExtensionsWizard } from './wizards/addExtensions/addExtensionsWizard';
 import { createTerminateDebugListener } from './wizards/debugging/terminateProcess';
@@ -41,14 +41,22 @@ export function activate(context: ExtensionContext) {
     })
   );
 
+  let updatedDocumentsCache: string[] = [];
   // When extension is started, loop for each text documents which are opened to update their language ID.
   workspace.textDocuments.forEach(document => {
-    updateLanguageId(document);
+    updateLanguageId(document, updatedDocumentsCache, false);
   });
   // When a text document is opened,  update their language ID.
   context.subscriptions.push(
     workspace.onDidOpenTextDocument((document) => {
-      updateLanguageId(document);
+      updateLanguageId(document, updatedDocumentsCache, true);
+    })
+  );
+  context.subscriptions.push(
+    workspace.onDidChangeConfiguration((e: ConfigurationChangeEvent) => {
+      if (e.affectsConfiguration(QuarkusConfig.QUARKUS_CONFIG_NAME)) {
+        updatedDocumentsCache = [];
+      }
     })
   );
 
@@ -56,8 +64,10 @@ export function activate(context: ExtensionContext) {
    * Update if required the language ID to 'quarkus-properties' if needed.
    *
    * @param document the text document.
+   * @param documentCache cache of documents for which the user has already been notified about the
+   * @param onExtensionLoad if the user manually changed the language id.
    */
-  async function updateLanguageId(document: TextDocument) {
+  async function updateLanguageId(document: TextDocument, documentCache: string[], onExtensionLoad: boolean) {
     const propertiesLanguageMismatch: PropertiesLanguageMismatch = QuarkusConfig.getPropertiesLanguageMismatch();
     if (propertiesLanguageMismatch === PropertiesLanguageMismatch.ignore) {
       // Do nothing
@@ -69,13 +79,13 @@ export function activate(context: ExtensionContext) {
         // the language ID is already quarkus-properties do nothing.
         return;
       }
-      tryToForceLanguageId(document, fileName, propertiesLanguageMismatch, 'quarkus-properties');
+      tryToForceLanguageId(document, fileName, documentCache, propertiesLanguageMismatch, 'quarkus-properties', onExtensionLoad);
     } else if (fileName === 'application.yaml' || fileName === 'application.yml') {
       if (document.languageId === 'yaml') {
         // the language ID is already yaml do nothing.
         return;
       }
-      tryToForceLanguageId(document, fileName, propertiesLanguageMismatch, 'yaml');
+      tryToForceLanguageId(document, fileName, documentCache, propertiesLanguageMismatch, 'yaml', onExtensionLoad);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -102,6 +102,16 @@ export function activate(context: ExtensionContext) {
         } else {
           // The application.properties file belong to a Quarkus project, force to the quarkus-properties language
           languages.setTextDocumentLanguage(document, languageId);
+          const CONFIGURE_IN_SETTINGS = "Configure";
```
fbricon: you want an option to not show it again

datho7561: I personally don't think that it's worthwhile to do this. The notification only appears when they try to manually change the language ID off of Quarkus properties:

LanguageIdUpdatedNotification

After they have been notified once that what they just did had no effect, the user would either change the setting to disable the automatic updating (which means they will never get the notification again) or not try the action again, since they know it will do nothing. As long as they don't try the action again, the notification shouldn't pop up again.

xorye:
The notification shouldn't appear when the languageId switch happens here:
https://github.com/redhat-developer/vscode-quarkus/blob/76f41f8ca4d0d6194c637bcaf3e4246be73cf3e9/src/extension.ts#L44-L47

since the check here happens right when vscode-quarkus first starts rather than when the user manually changes the languageId.

datho7561:
> The notification shouldn't appear when the languageId switch happens here:

I pushed an amend which handles this. I checked it manually by disabling the extension, changing the language on the properties file to something else, then restarting the extension. The language was automatically switched, but no notification showed up.

src/extension.ts
datho7561: I added a "Disable Language Updating" button to the dialogue and changed the configure button to "Configure in Settings"

src/extension.ts
```diff
@@ -57,7 +57,7 @@ export function activate(context: ExtensionContext) {
    *
    * @param document the text document.
```
xorye, line 66: Could we add documentation for `@param userInitiated`

src/extension.ts
```diff
@@ -85,7 +85,7 @@ export function activate(context: ExtensionContext) {
...
    * @param languageId the language ID.
```
xorye, line 97: Same here for `@param userInitiated` please

src/extension.ts
```diff
@@ -101,7 +101,25 @@ export function activate(context: ExtensionContext) {
...
+            const DISABLE_IN_SETTINGS = "Disable Language Updating";
+            const response: Thenable<string> = window.showInformationMessage(
+                `Quarkus Tools for Visual Studio Code automatically switched the language ID of '${fileName}' `
+                + `to be '${languageId}' in order to provide language support. `
+                + `This behavior can be configured in settings.`, DISABLE_IN_SETTINGS, CONFIGURE_IN_SETTINGS);
+            response.then(result => {
+              if (result === CONFIGURE_IN_SETTINGS) {
+                commands.executeCommand('workbench.action.openSettings', QuarkusConfig.PROPERTIES_LANGUAGE_MISMATCH);
+              } else if (result === DISABLE_IN_SETTINGS) {
+                workspace.getConfiguration().update(QuarkusConfig.PROPERTIES_LANGUAGE_MISMATCH, PropertiesLanguageMismatch.ignore).then(() => {
```
xorye: Could we create a method in `QuarkusConfig` like:
```
  export function setPropertiesLanguageMismatch(value: PropertiesLanguageMismatch): Thenable<void> {
    return saveToQuarkusConfig<PropertiesLanguageMismatch>(PROPERTIES_LANGUAGE_MISMATCH, value);
  }
```

and use it here?

`saveToQuarkusConfig` would also need to be changed to return the thenable:
```
  export function saveToQuarkusConfig<T>(configName: string, value: T) {
    return workspace.getConfiguration().update(configName, value, ConfigurationTarget.Global);
  }
```

The main reason for `QuarkusConfig` was to make it easier to maintain correct typing, ie, we don't want to accidentally set `QuarkusConfig.PROPERTIES_LANGUAGE_MISMATCH` to a boolean value etc.

src/extension.ts
```diff
@@ -85,7 +85,7 @@ export function activate(context: ExtensionContext) {
    * @param document the text document.
    * @param languageId the language ID.
    */
-  async function tryToForceLanguageId(document: TextDocument, fileName: string, propertiesLanguageMismatch: PropertiesLanguageMismatch, languageId: string) {
+  async function tryToForceLanguageId(document: TextDocument, fileName: string, propertiesLanguageMismatch: PropertiesLanguageMismatch, languageId: string, userInitiated: boolean) {
```
xorye: Not a problem with this PR, but we should remove the `async` modifier since this function does not use `await` anywhere in the body.

datho7561: Okay. I think the original intent may have been to show it returns a promise, so I added a `Promise<void>` return annotation.

xorye:
I found a weird interaction when `quarkus.tools.propertiesLanguageMismatch` is `forceQuarkus` and after I set the languageId for the properties file to C:

notification

Since its set to `forceQuarkus`, the notification pops up informing the user about the forcing. The languageId is set to `quarkus-properties`. But when I go to a different tab and click back to application.properties, for a split second on the bottom right, the languageId is still C for some reason. It should not be C, since it was forced to `quarkus-properties`. Not sure why this is happening. 

It has the languageId as C for a split second, and then it forces it to `quarkus-properties` again, therefore showing the notification again.

xorye:
A similar thing happens when `quarkus.tools.propertiesLanguageMismatch` is set to `prompt`:
prompt

After clicking `Yes`, the file should stay as a quarkus-properties file, but it does not

datho7561:
The same thing happens when you set it to something else, let it get automatically updated, disable the updating, then close and reopen the file. The file is validated as whatever you set it to instead of a Quarkus properties file.

xorye:
> The same thing happens when you set it to something else, let it get automatically updated, disable the updating, then close and reopen the file. 

I'm not sure if I completely understood, but I think this is a default VS Code feature? When a file is completely closed and opened again, it seems like VS Code just sets the languageId to whatever seems appropriate.

My issue is specific for when we go to different tab(s) and come back to application.properties. 

Related issue: https://github.com/microsoft/vscode/issues/103198

datho7561:
> When a file is completely closed and opened again, it seems like VS Code just sets the languageId to whatever seems appropriate.

I see this behaviour now. I'll look into the API to see if there are flags that indicate if the change was done manually by the user.

datho7561:
> I see this behaviour now. I'll look into the API to see if there are flags that indicate if the change was done manually by the user.

Doesn't look like there is anything that can help differentiate a manual user change from any built-in automatic changes in the text document `languageId`. This also might be the intended behaviour, since the user sets their "preferred language" by changing it manually, and VSCode defaults to opening the document with the preferred language.

datho7561:
In the latest amend, I use a cache of which documents the user has been prompted about in order to avoid re-prompting the user about this setting.

UsingADocumentCache

PR #300, opened 2020-08-20
"Add extensions" and "Debug" commands prompt for standard mode

### LINKED ISSUE #299: Debug quarkus project command doesn't appear on Quarkus projects after going from lightweight mode to standard mode
When the Java language server starts in lightweight mode, the project label info command (which is called when vscode-quarkus first starts) fails. Therefore the debug current quarkus project command (and add extensions command) will not appear in the command palette.

When the Java language server is switched to standard mode we still do not have those commands because the project labels command isn't triggered. The user would have to trigger it by adding or removing a workspace file: 
https://github.com/redhat-developer/vscode-quarkus/blob/f38f4caaf218cf9c6ce91e64a0d9cd632314a483/src/QuarkusProjectListener.ts#L38

datho7561:
Add extensions and Debug are always visible in the command palette. If either Add extensions or Debug is selected while the Java server is in LightWeight mode, the user is prompted with a dialog that allows them to switch to Standard mode. The actions are performed after the switch is made.

This solution was adopted from [a vscode-java-debug commit that solves a similar issue](https://github.com/microsoft/vscode-java-debug/commit/20cd68cbb112f170b4b0aaf849604cf37853091a)

Fixes #299

### DIFF
--- a/package.json
+++ b/package.json
@@ -190,11 +190,11 @@
       "commandPalette": [
         {
           "command": "quarkusTools.addExtension",
-          "when": "quarkusProjectExists"
+          "when": "quarkusProjectExistsOrLightWeight"
         },
         {
           "command": "quarkusTools.debugQuarkusProject",
-          "when": "quarkusProjectExists"
+          "when": "quarkusProjectExistsOrLightWeight"
         }
       ]
     },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,7 +15,7 @@
  */
 import * as path from 'path';
 import { VSCodeCommands } from './definitions/constants';
-import { ExtensionContext, commands, window, workspace, Terminal, languages, TextDocument, ConfigurationChangeEvent } from 'vscode';
+import { ExtensionContext, commands, window, workspace, Terminal, languages, TextDocument, ConfigurationChangeEvent, Disposable } from 'vscode';
 import { QuarkusContext } from './QuarkusContext';
 import { addExtensionsWizard } from './wizards/addExtensions/addExtensionsWizard';
 import { createTerminateDebugListener } from './wizards/debugging/terminateProcess';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -186,18 +186,6 @@
         }
       }
     },
-    "menus": {
-      "commandPalette": [
-        {
-          "command": "quarkusTools.addExtension",
-          "when": "quarkusProjectExists"
```
angelozerr, line 193: If you remove this declaration where `quarkusProjectExists` condition is used?

src/extension.ts
```diff
@@ -131,14 +132,22 @@ function registerVSCodeCommands(context: ExtensionContext) {
    * Command for adding Quarkus extensions to current Quarkus Maven project
    */
   context.subscriptions.push(commands.registerCommand(VSCodeCommands.ADD_EXTENSIONS, () => {
-    addExtensionsWizard();
+    requestStandardMode("Adding extensions").then((isStandardMode) => {
+      if (isStandardMode) {
+        addExtensionsWizard();
```
angelozerr: I think it misses a test with `quarkusProjectExists` context

src/extension.ts
```diff
@@ -131,14 +132,22 @@ function registerVSCodeCommands(context: ExtensionContext) {
...
+    });
   }));
 
   /**
    * Command for debugging current Quarkus Maven project
    */
   context.subscriptions.push(commands.registerCommand(VSCodeCommands.DEBUG_QUARKUS_PROJECT, () => {
-    tryStartDebugging();
+    requestStandardMode("Debugging the project").then((isStandardMode) => {
+      if (isStandardMode) {
```
angelozerr, line 190: I think it misses a test with `quarkusProjectExists` context

datho7561:
Thanks for the review Angelo. After looking into what you mentioned, this PR is pretty broken, and needs a lot more work that will take a while to complete.

datho7561:
~THIS PR IS NOT READY~ Fixed, ignore the rest of this comment
---
I am having trouble getting the menu options to conditionally show up, since if I use the `"when"` clause, they don't appear in the command palette before the extension is activated. Right now, this PR just gives an error message when the commands are used instead of hiding the commands.

Personally, I think that this has more drawbacks then benefits.

DEMO with a Helidon project:
QuarkusDebugAndAddExtension

datho7561:
To test, open a Quarkus project with vscode-java in lightweight mode. Add extension should appear in the command palette. Switch to standard mode. Add extension should still be in the command palette. If you try to use add extension in lightweight mode, it will prompt to switch to standard mode, then will open the add extension menu when vscode-java is in standard mode.

Then, do the same for a non-Quarkus MicroProfile project (Helidon, OpenLiberty, etc.). Add extension appears in lightweight mode, but not standard. If you try to use add extension in lightweight mode, it will switch vscode-java to standard mode, then give an error stating this is not a Quarkus project.

PR #310, opened 2020-10-09
Better UX when generating a Quarkus project

### LINKED ISSUE #309: Better UX when generating a Quarkus project
Hello,
I'd like to propose a UX enhancement of generating project wizard.

### Current behavior
Currently, on Fedora with GNOME DE, it opens a dialog with a list of recent folders
image

When running in Eclipse Che, it opens a dialog pointed to the system root folder
image

In both cases, the user needs to navigate to some folder that is most likely far from the initially proposed one.

This is a default behavior as it [opens a dialog](https://github.com/redhat-developer/vscode-quarkus/blob/4c4f56ff55b5912e5f77ee6a861e0683eab24731/src/wizards/generateProject/generationWizard.ts#L198) without specifying an [initial directory](https://github.com/eclipse-theia/theia/blob/eb1136b1180e63312b01636f5ff65dce24245441/packages/plugin/src/theia.d.ts#L2648-L2651).

### Proposed solution
I want to propose setting the 1'st workspace folder returned by [`workspace.workspaceFolders` API](https://github.com/eclipse-theia/theia/blob/eb1136b1180e63312b01636f5ff65dce24245441/packages/plugin/src/theia.d.ts#L4928-L4934) as an initial directory.

The opened dialog will point to a directory that is **much closer** to the user's currently opened project and it will require less navigating through the file system.

azatsarynnyy:
Signed-off-by: Artem Zatsarynnyi <azatsary@redhat.com>

closes https://github.com/redhat-developer/vscode-quarkus/issues/309
also closes downstream Che issue https://github.com/eclipse/che/issues/18006

### DIFF
--- a/src/wizards/generateProject/generationWizard.ts
+++ b/src/wizards/generateProject/generationWizard.ts
@@ -167,12 +167,13 @@ async function getTargetDirectory(projectName: string) {
   const OPTION_OVERWRITE = 'Overwrite';
   const OPTION_CHOOSE_NEW_DIR = 'Choose new directory';
 
-  let directory: Uri|undefined = await showOpenFolderDialog({ openLabel: LABEL_CHOOSE_FOLDER });
+  const defaultDirectoryUri = workspace.workspaceFolders ? workspace.workspaceFolders[0].uri : undefined;
+  let directory: Uri|undefined = await showOpenFolderDialog({ openLabel: LABEL_CHOOSE_FOLDER, defaultUri: defaultDirectoryUri});
 
   while (directory && fs.existsSync(path.join(directory.fsPath, projectName))) {
     const overrideChoice: string = await window.showWarningMessage(MESSAGE_EXISTING_FOLDER, OPTION_OVERWRITE, OPTION_CHOOSE_NEW_DIR);
     if (overrideChoice === OPTION_CHOOSE_NEW_DIR) {
-      directory = await showOpenFolderDialog({ openLabel: LABEL_CHOOSE_FOLDER });
+      directory = await showOpenFolderDialog({ openLabel: LABEL_CHOOSE_FOLDER, defaultUri: defaultDirectoryUri});
     } else if (overrideChoice === OPTION_OVERWRITE) {
       break;
     } else { // User closed the warning window


src/wizards/generateProject/generationWizard.ts
```diff
@@ -167,12 +167,13 @@ async function getTargetDirectory(projectName: string) {
   const OPTION_OVERWRITE = 'Overwrite';
   const OPTION_CHOOSE_NEW_DIR = 'Choose new directory';
 
-  let directory: Uri|undefined = await showOpenFolderDialog({ openLabel: LABEL_CHOOSE_FOLDER });
+  const defaultDirectoryUri = workspace.workspaceFolders ? workspace.workspaceFolders[0].uri : undefined;
```
rgrunber, line 170: Could this use some kind of guard? Under Theia (with this change), if I attempt to run the wizard when no workspace has been selected, I get an exception.

Maybe something like (array && array.length) is needed.

rgrunber (commented):
The change definitely improves the situation for the wizard creation. Much nicer to have the current workspace folder. Just one minor issue I found.

src/wizards/generateProject/generationWizard.ts
tsmaeder: Should probably be `(workspaceFolders && workspaceFolders.length > 0) ?....`

azatsarynnyy: Thank you @rgrunber for checking it.
I believe no additional guard is needed that would be a Theia-specific one.
`workspace.workspaceFolders` API must behave the same in VS Code and Theia.
[The doc says](https://github.com/eclipse-theia/theia/blob/c4f324ada78e8074c401fa52a0a655bbb64de017/packages/plugin/src/theia.d.ts#L5041-L5047) it must return `undefined` when no folder is open.

I've just filed a Theia bug https://github.com/eclipse-theia/theia/issues/8640
Working on fixing it. I'll let you know when it's fixed.

fbricon:
Ideally, we should even preselect the last opened (parent) folder, where a previous project was generated.

src/wizards/generateProject/generationWizard.ts
azatsarynnyy: @rgrunber I've fixed the bug in Theia https://github.com/eclipse-theia/theia/pull/8641 - it works identically to VS Code.

PR #316, opened 2020-10-30
Deploy to OpenShift command

### LINKED ISSUE #313: Provide a "Deploy to OpenShift" command
Here's the idea:
- if there's at least one Quarkus project in the workspace, then enable a "Deploy to OpenShift" command
- when invoked, the command will:
  - detect if vscode-openshift-tools is installed, if not, propose the user to install it
  - ask the user to select the quarkus project to deploy, if there are several.
  - invoke some (new) vscode-openshift-tools API/command, requesting to deploy the selected project 
  - see how to handle already published projects (then re-push them?)

On the vscode-openshift-tools side, it'll be equivalent to selecting a workspace directory deployment.
cc @dgolovin @mohitsuman

datho7561:
Uses the existing menu to select which Quarkus project to deploy.
Uses OpenShift Connector in order to deploy the Quarkus project to
OpenShift.
OpenShift Connector handles selecting/creating the application and
naming the component.

Closes #313

### DIFF
--- a/src/utils/openShiftConnectorUtils.ts
+++ b/src/utils/openShiftConnectorUtils.ts
@@ -1,0 +1,97 @@
+/**
+ * Copyright 2021 Red Hat, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { commands, Disposable, Extension, extensions, ProgressLocation, Uri, window } from "vscode";
+
+export const OPENSHIFT_CONNECTOR_EXTENSION_ID = 'redhat.vscode-openshift-connector';
+export const OPENSHIFT_CONNECTOR = 'OpenShift Connector extension';
+const DOWNLOAD_TIMEOUT = 60000; // Timeout for downloading VSCode OpenShift Connector, in milliseconds
+
+/**
+ * Returns true if the OpenShift connector extension is installed, and false otherwise
+ *
+ * @returns true if the OpenShift connector extension is installed, and false otherwise
+ */
+export function isOpenShiftConnectorInstalled(): boolean {
+  return !!extensions.getExtension(OPENSHIFT_CONNECTOR_EXTENSION_ID);
+}
+
+/**
+ * Returns the OpenShift Connector extension API
+ *
+ * @throws Error if the extension is not installed
+ * @returns the OpenShift Connector extension API
+ */
+export async function getOpenShiftConnector(): Promise<any> {
+  if (!isOpenShiftConnectorInstalled()) {
+    throw new Error(`${OPENSHIFT_CONNECTOR} is not installed`);
+  }
+  const openShiftConnector: Extension<any> = extensions.getExtension(OPENSHIFT_CONNECTOR_EXTENSION_ID);
+  if (openShiftConnector.isActive) {
+    return openShiftConnector.exports;
+  }
+  return extensions.getExtension(OPENSHIFT_CONNECTOR_EXTENSION_ID).activate();
+}
+
+/**
+ * Install the OpenShift Connector extension
+ *
+ * @returns when the extension is installed
+ * @throws if the user refuses to install the extension, or if the extension does not get installed within a timeout period
+ */
+async function installOpenShiftConnector(): Promise<void> {
+  let installListenerDisposable: Disposable;
+  return new Promise<void>((resolve, reject) => {
+    installListenerDisposable = extensions.onDidChange(() => {
+      if (isOpenShiftConnectorInstalled()) {
+        resolve();
+      }
+    });
+    commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID)
+      .then((_unused: any) => { }, reject);
+    setTimeout(reject, DOWNLOAD_TIMEOUT, new Error(`${OPENSHIFT_CONNECTOR} installation is taking a while. Cancelling 'Deploy to OpenShift'. Please retry after the OpenShift Connector installation has finished`));
+  }).finally(() => {
+    installListenerDisposable.dispose();
+  });
+}
+
+/**
+ * Install the OpenShift Connector extension and show the progress
+ *
+ * @returns when the extension is installed
+ * @throws if the extension installation fails or times out
+ */
+export async function installOpenShiftConnectorWithProgress(): Promise<void> {
+  await window.withProgress({ location: ProgressLocation.Notification, title: `Installing ${OPENSHIFT_CONNECTOR}...` }, progress => {
+    const openShiftConnectorInstall: Promise<void> = installOpenShiftConnector();
+    openShiftConnectorInstall.catch((e) => {
+      window.showErrorMessage(`${e}`);
+    });
+    return openShiftConnectorInstall;
+  });
+  window.showInformationMessage(`Successfully installed ${OPENSHIFT_CONNECTOR}.`);
+}
+
+/**
+ * Create a component from the given Quarkus project, then push the component to OpenShift
+ *
+ * @param projectUri the uri of the Quarkus project to deploy
+ */
+export async function deployQuarkusProject(projectUri: Uri): Promise<void> {
+  await commands.executeCommand("openshift.component.deployRootWorkspaceFolder", //
+    projectUri,
+    "java-quarkus"
+  );
+}
--- a/src/wizards/deployToOpenShift/deployToOpenShift.ts
+++ b/src/wizards/deployToOpenShift/deployToOpenShift.ts
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2021 Red Hat, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Uri, window } from "vscode";
+import { ProjectLabelInfo } from "../../definitions/ProjectLabelInfo";
+import { deployQuarkusProject, getOpenShiftConnector, installOpenShiftConnectorWithProgress, isOpenShiftConnectorInstalled, OPENSHIFT_CONNECTOR } from "../../utils/openShiftConnectorUtils";
+import { getQuarkusProject } from "../getQuarkusProject";
+
+/**
+ * Attempts to select a Quarkus project, then deploy it to OpenShift using OpenShift Connector
+ *
+ * @returns when the deploy to OpenShift succeeds, or when the deploy to OpenShift fails
+ * @throws if the OpenShift Connector installation fails
+ */
+export async function deployToOpenShift(): Promise<void> {
+  await installOpenShiftConnectorIfNeeded();
+  const quarkusProject: ProjectLabelInfo = await getQuarkusProject();
+  const quarkusProjectUri: Uri = Uri.parse(quarkusProject.uri);
+  await deployQuarkusProject(quarkusProjectUri);
+}
+
+/**
+ * Installs the OpenShift Connector extension if its missing, then returns the extension API
+ *
+ * @throws Error if the extension is not installed and something prevents installation
+ * @returns the OpenShift Connector extension API
+ */
+async function installOpenShiftConnectorIfNeeded(): Promise<any> {
+  if (isOpenShiftConnectorInstalled()) {
+    return getOpenShiftConnector();
+  }
+  return askToInstallOpenShiftConnector();
+}
+
+/**
+ * Resolves to the OpenShift connector extension API, or rejects if the extension is missing and the user doesn't want to install it
+ *
+ * @throws Error if the extension is missing and the user can't or doesn't want to install it
+ * @returns the OpenShift connector extension API
+ */
+async function askToInstallOpenShiftConnector(): Promise<any> {
+  const YES: string = 'Yes';
+  const NO: string = 'No';
+  const response: string = await window.showInformationMessage(`${OPENSHIFT_CONNECTOR} is needed to deploy to OpenShift. Install it now?`, YES, NO);
+  if (response === YES) {
+    try {
+      await installOpenShiftConnectorWithProgress();
+      if (isOpenShiftConnectorInstalled()) {
+        return getOpenShiftConnector();
+      }
+    } catch (e) {
+      throw new Error(`${OPENSHIFT_CONNECTOR} installation failed`);
+    }
+  }
+  throw new Error(`${OPENSHIFT_CONNECTOR} needs to be installed to deploy to OpenShift`);
+}
--- a/.vscode/snippets.code-snippets
+++ b/.vscode/snippets.code-snippets
@@ -1,0 +1,23 @@
+{
+  "Copyright Header": {
+    "scope": "javascript,typescript",
+    "prefix": "copyright",
+    "body": [
+      "/**",
+      " * Copyright ${CURRENT_YEAR} ${1:Red Hat, Inc.} and others.",
+      " *",
+      " * Licensed under the Apache License, Version 2.0 (the \"License\");",
+      " * you may not use this file except in compliance with the License.",
+      " * You may obtain a copy of the License at",
+      " *",
+      " *     http://www.apache.org/licenses/LICENSE-2.0",
+      " *",
+      " * Unless required by applicable law or agreed to in writing, software",
+      " * distributed under the License is distributed on an \"AS IS\" BASIS,",
+      " * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.",
+      " * See the License for the specific language governing permissions and",
+      " * limitations under the License.",
+      " */"
+    ]
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,58 +25,58 @@
       }
     },
     "@nut-tree/libnut": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/@nut-tree/libnut/-/libnut-1.0.2.tgz",
-      "integrity": "sha512-XwwMYVqY4Lo6i45f5gMy3IF5V0njwxIsLqtWbn1s1sKGMAz61zP4xFytgLxSgFQSukSvoQeawtBWq/WlNvtTmQ==",
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@nut-tree/libnut/-/libnut-2.1.0.tgz",
+      "integrity": "sha512-aCSbURPPYC6hwJG9lT9JsjWwC8jj4Hpv2CoGIIDh+l4oPvIPTLvGTabOkcmEgz5zIxmUWAJ8kCZsnxsGE4B/EA==",
       "dev": true,
       "requires": {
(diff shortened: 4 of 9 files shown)


datho7561:
@mohitsuman @dgolovin This is a POC for selecting a Quarkus project to deploy from the current workspace. The next step would be integrating with OpenShift Connector in order to deploy the selected Quarkus project to OpenShift.

src/wizards/deployToOpenShift/deployToOpenShift.ts
```diff
@@ -0,0 +1,51 @@
...
+import { getQuarkusProject } from "../getQuarkusProject";
+
+/**
+ * Attempts to select a Quarkus project, then deploy it to OpenShift using OpenShift Connector
+ */
+export async function deployToOpenShift() {
+  try {
+    const openShiftConnectorExtension: any = await installOpenShiftConnectorIfNeeded();
+    const quarkusProject: ProjectLabelInfo = await getQuarkusProject();
+    // openShiftConnectorExtension.Component.createFromFolder(quarkusProject.uri);
```
dgolovin: Considering we don't have API for that yet, I would suggest to use vscode 'commands' API. Here is the issue for extension to provide vscode command for that https://github.com/redhat-developer/vscode-openshift-tools/issues/1832.

datho7561:
I think this should work now. I don't have access to an OpenShift cluster to test this on, though. I tried with minikube and it looks like that doesn't work.

How to use:
 * Build and install a .vsix of [OpenShift Connector](https://github.com/redhat-developer/vscode-openshift-tools) from the master branch on the GitHub
 * Install the vscode-microprofile extension from your preferred extension marketplace
 * Follow [these instructions](https://github.com/redhat-developer/vscode-quarkus/blob/master/CONTRIBUTING.md#development-setup) to get a development environment for this extension
 * Build and run this extension
 * Make a Quarkus project
 * Open the folder that contains the Quarkus project
 * Run the "Quarkus: Deploy to OpenShift" command

This should also work for multi root workspaces. There should be a dialog that prompts to select which project to deploy.

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+  return extensions.getExtension(OPENSHIFT_CONNECTOR_EXTENSION_ID).activate();
+}
+
+/**
+ * Install the OpenShift Connector extension
+ *
+ * Resolves when the extension is installed
+ */
+export async function installOpenShiftConnector(): Promise<any> {
+  commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID);
```
fbricon, line 37: return

fbricon (changes requested):
Didn't work for me:

Screenshot 2020-11-13 at 10 40 15

Maybe it's because I'm running vscode-quarkus from the Extension Development Host? The openshift extension does install though

src/utils/openShiftConnectorUtils.ts
fbricon:
```typescript
export async function installOpenShiftConnector(): Promise<any> {
  return window.withProgress({ location: ProgressLocation.Notification, title: `Installing ${OPENSHIFT_CONNECTOR}...`}, progress => {
    return commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID);
  }).then(() => {
    if (!!extensions.getExtension(OPENSHIFT_CONNECTOR_EXTENSION_ID)) {
      window.showInformationMessage(`Successfully installed ${OPENSHIFT_CONNECTOR}.`);
    } else {
      window.showErrorMessage(`Failed to install ${OPENSHIFT_CONNECTOR}.`);
    }
  });
```

fbricon: @jrieken any idea why calling extensions.getExtension(...) after calling "workbench.extensions.installExtension" always fails?

fbricon: Ok so my guess is we need to add a listener to extensions.onDidChange, once we detect the OPENSHIFT_CONNECTOR_EXTENSION_ID has been registered, then we can proceed with calling the openshift deployment command

fbricon:
After installing the latest vscode-openshift-tools CI build from https://download.jboss.org/jbosstools/adapters/snapshots/vscode-openshift-tools/?C=M;O=D, without even creating/connecting to an openshift cluster, the deploy command gives me 2 conflicting messages:

Screenshot 2020-11-13 at 17 44 00

@dgolovin if there's no cluster available, we want the deploy command to initiate the creation/configuration of the cluster

src/wizards/deployToOpenShift/deployToOpenShift.ts
```diff
@@ -0,0 +1,51 @@
...
+import { getQuarkusProject } from "../getQuarkusProject";
+
+/**
+ * Attempts to select a Quarkus project, then deploy it to OpenShift using OpenShift Connector
+ */
+export async function deployToOpenShift() {
+  try {
+    await installOpenShiftConnectorIfNeeded();
+    const quarkusProject: ProjectLabelInfo = await getQuarkusProject();
+    await commands.executeCommand("openshift.component.createFromRootWorkspaceFolder", quarkusProject.uri, "java-quarkus");
```
fbricon: Once a cluster is already available (or else it [fails](https://github.com/redhat-developer/vscode-openshift-tools/issues/1849)), the command lets you create an [application](https://github.com/redhat-developer/vscode-openshift-tools/issues/1850) and a component, if needed.

Then it fails with:

Screenshot 2020-11-13 at 18 03 23

Either the command call is missing something, or openshift-tools has an issue. @dgolovin which is it?

src/utils/openShiftConnectorUtils.ts
datho7561: I improved the error reporting on failed extension installation. It properly waits for OpenShift Connector to finish installing. I also added a timeout for the installation handling code.

datho7561:
> After installing the latest vscode-openshift-tools CI build from https://download.jboss.org/jbosstools/adapters/snapshots/vscode-openshift-tools/?C=M;O=D, without even creating/connecting to an openshift cluster, the deploy command gives me 2 conflicting messages:

I was expecting the command to reject if something goes wrong. If there is a command that returns if there is a cluster or rejects if there is none, I could use that before trying to deploy.

datho7561:
> After installing the latest vscode-openshift-tools CI build from https://download.jboss.org/jbosstools/adapters/snapshots/vscode-openshift-tools/?C=M;O=D, without even creating/connecting to an openshift cluster, the deploy command gives me 2 conflicting messages:

I was expecting the command to reject if something goes wrong. If there is a command that checks the status of the cluster, I could use that before trying to deploy.

src/wizards/deployToOpenShift/deployToOpenShift.ts
dgolovin: @fbricon will look at this, it could be odo failure, bz project already has devfile.yaml. I am going to fix it by detecting project with devfile.yaml in root folder and add option `--devfile devfile.yaml`, so it will just use existing file.

dgolovin:
@datho7561 @fbricon I generated quarkus project, it is using maven. Then I am trying to deploy and get this error: "No quarkus project detected in **this folder**". How do I set **this folder**? 
quarkus-deploy-project

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+        commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID)
+          .then((_unused: any) => { }, reject);
+      }),
+      new Promise((_unused, reject) => {
+        // Fail install if it takes longer than the timeout period
+        setTimeout(reject, DOWNLOAD_TIMEOUT, new Error(`${OPENSHIFT_CONNECTOR} installation is taking a while. Cancelling 'Deploy to OpenShift'`));
+      })
+    ]);
+  } catch (e) {
+    installDisposables.forEach((d: Disposable) => { d.dispose(); });
```
dgolovin: The code creates Promise instance and returns it. This catch block is basically dead branch IMO, it will never get control. Exceptions inside callbacks will be caught at line 50 in `deployToOpenShift.ts`.

datho7561: Good catch, thanks! I'll fix this later today

datho7561:
@fbricon it looks like the issue dgolovin ran into is related to jdt.ls not automatically importing the project. I'll investigate and open a bug this afternoon.

@dgolovin If you open the project folder (`quarkus-getting-started`) through `Ctrl+K O`, it will complete some initial setup that should have been done automatically. Then, if you try the Deploy to OpenShift, it should work.

I'll also ~~open an issue for improving~~ improve the "no projects detected" to better explain which folders it looks through for the Quarkus project.

src/wizards/deployToOpenShift/deployToOpenShift.ts
dgolovin: @fbricon https://github.com/redhat-developer/vscode-openshift-tools/issues/1815 is fixed. Released to master branch.

datho7561: @dgolovin Thanks for the update! I looked over the code, and it seems I was just calling the command with the wrong parameters. It should work now. I'll record a short demo of the new workflow.

src/wizards/deployToOpenShift/deployToOpenShift.ts
```diff
@@ -0,0 +1,61 @@
...
+import { getQuarkusProject } from "../getQuarkusProject";
+
+/**
+ * Attempts to select a Quarkus project, then deploy it to OpenShift using OpenShift Connector
+ */
+export async function deployToOpenShift() {
+  try {
+    await installOpenShiftConnectorIfNeeded();
+    const quarkusProject: ProjectLabelInfo = await getQuarkusProject();
+    const result: string = await commands.executeCommand("openshift.component.createFromRootWorkspaceFolder", //
```
datho7561: I am looking into trying to push the component, then activating the "create component" if that fails. I am hoping this will cover the use case where there is a `devfile.yaml` and a `.odo` folder, but the component is missing from OpenShift. I'm also going to wrap the command calls to a different interface to make this function cleaner to read.

src/wizards/deployToOpenShift/deployToOpenShift.ts
```diff
@@ -0,0 +1,61 @@
...
+      Uri.parse(quarkusProject.uri) as Uri, //
+      [] as Uri[], // unused in OpenShift Connector
+      undefined, // OpenShift context
+      "java-quarkus");
+    if (!result) {
+      throw new Error();
+    }
+    return window.showInformationMessage(result);
+  } catch (error) {
+    return window.showErrorMessage(`Failed to deploy to OpenShift`);
```
datho7561: I need to remove the error/success reporting here, since OpenShift connector will handle this

src/wizards/deployToOpenShift/deployToOpenShift.ts
datho7561: I changed strategies to checking the folder for `.odo` to detect if the component should be created or just pushed.

dgolovin (approved):
Except the fact that some `return Promise.resolve();` and some try/catch blocks are not required,  It looks good to me. Haven't tried it though.

src/wizards/deployToOpenShift/deployToOpenShift.ts
```diff
@@ -0,0 +1,65 @@
+import { Uri, window } from "vscode";
```
datho7561, line 16: If you try to deploy a Quarkus project, but aren't connected to a cluster, then the error message is not friendly.

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+      commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID)
+        .then((_unused: any) => { }, reject);
+    }),
+    new Promise<void>((_unused, reject) => {
+      // Fail install if it takes longer than the timeout period
+      setTimeout(reject, DOWNLOAD_TIMEOUT, new Error(`${OPENSHIFT_CONNECTOR} installation is taking a while. Cancelling 'Deploy to OpenShift'`));
+    })
+  ]).catch(e => {
+    installListenerDisposables.forEach((d: Disposable) => { d.dispose(); });
+    return Promise.reject(e);
```
datho7561: `throw e`

src/wizards/deployToOpenShift/deployToOpenShift.ts
datho7561: I don't seem to get any error message, but the dialog to push doesn't show up if there is no cluster connected.

datho7561: If you create the component when the cluster is not connected, you do not get validation on the project field, so it could be easy to get the config wrong. It might make sense to only allow deploying to OpenShift through this command when vscode-openshift is able to connect to a cluster.

datho7561: I implemented a hack where it runs `kubectl api-resources` to see if the computer is connected to a cluster. If not connected, then it doesn't try to create the component.

fbricon:
It should be openshift-tools' responsibility to perform the  necessary checks for deploying to openshift (stuff involving kubectl for instance). @dgolovin please check what parts of this PR should be moved to openshift tools

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/**
+ * Install the OpenShift Connector extension
+ *
+ * @returns when the extension is installed
+ * @throws if the user refuses to install the extension, or if the extension does not get installed within a timeout period
+ */
+async function installOpenShiftConnector(): Promise<void> {
+  const installListenerDisposables: Disposable[] = [];
```
dgolovin: Should be just `installListenerDisposable: Disposable;` because it is always one returned from `onDidChange` call. That will reduce `forEach` to just single call of `dispose()`.

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+        }
+      }, null, installListenerDisposables);
+      commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID)
+        .then((_unused: any) => { }, reject);
+    }),
+    new Promise<void>((_unused, reject) => {
+      // Fail install if it takes longer than the timeout period
+      setTimeout(reject, DOWNLOAD_TIMEOUT, new Error(`${OPENSHIFT_CONNECTOR} installation is taking a while. Cancelling 'Deploy to OpenShift'`));
+    })
+  ]).catch(e => {
```
dgolovin: should be
```
]).finally(() => {
    installListenerDisposable.dispose();
})
```

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Install the OpenShift Connector extension
+ *
+ * @returns when the extension is installed
+ * @throws if the user refuses to install the extension, or if the extension does not get installed within a timeout period
+ */
+async function installOpenShiftConnector(): Promise<void> {
+  const installListenerDisposables: Disposable[] = [];
+  return Promise.race([
+    new Promise<void>((resolve, reject) => {
+      extensions.onDidChange(() => {
```
dgolovin: ```installListenerDisposable = extensions.onDidChange(() => {```

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+        if (isOpenShiftConnectorInstalled()) {
+          installListenerDisposables.forEach((d: Disposable) => { d.dispose(); });
```
dgolovin: This line is not needed when using `finally()` see change below.

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+          resolve();
+        }
+      }, null, installListenerDisposables);
```
dgolovin: These parameters are not needed.

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+      commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID)
+        .then((_unused: any) => { }, reject);
+    }),
+    new Promise<void>((_unused, reject) => {
+      // Fail install if it takes longer than the timeout period
+      setTimeout(reject, DOWNLOAD_TIMEOUT, new Error(`${OPENSHIFT_CONNECTOR} installation is taking a while. Cancelling 'Deploy to OpenShift'`));
+    })
+  ]).catch(e => {
+    installListenerDisposables.forEach((d: Disposable) => { d.dispose(); });
+    throw e;
```
dgolovin: No need to re-throw the exception.

src/wizards/deployToOpenShift/deployToOpenShift.ts
```diff
@@ -0,0 +1,66 @@
...
+  const NO: string = 'No';
+  const response: string = await window.showInformationMessage(`${OPENSHIFT_CONNECTOR} is needed to deploy to OpenShift. Install it now?`, YES, NO);
+  if (response === YES) {
+    try {
+      await installOpenShiftConnectorWithProgress();
+      if (isOpenShiftConnectorInstalled()) {
+        return getOpenShiftConnector();
+      }
+    } catch (e) {
+      throw new Error(`${OPENSHIFT_CONNECTOR} installation failed`);
```
dgolovin, line 64: I am running vscode-quarkus from dev env. Whenever I simulate this exception the error is only reported in console, it is never reported to the user through `window.showErrorMessage`.

datho7561:
Thanks for the review, Denis! I have addressed the comments.

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,150 @@
...
+  return extensions.getExtension(OPENSHIFT_CONNECTOR_EXTENSION_ID).activate();
+}
+
+/**
+ * Install the OpenShift Connector extension
+ *
+ * @returns when the extension is installed
+ * @throws if the user refuses to install the extension, or if the extension does not get installed within a timeout period
+ */
+async function installOpenShiftConnector(): Promise<void> {
```
dgolovin, line 54: Even better version for this function.
Notice: 
1. no Promise.race() required anymore
2. no unused parameters in callback
3. Promise.resolve(commands.executeCommand()) call converts `thenable` instance to `Promise` and that in turn allows to use just `.catch(reject)`.
4. Even when timer expired after Promise fulfilled, rejecting it won't have any effect on fulfilled promise
5. ?. is not required because it `installListenerDisposable` is always going to be initialized
``` javascript
async function installOpenShiftConnector(): Promise<void> {
  let installListenerDisposable: Disposable;

  return new Promise<void>((resolve, reject) => {
    installListenerDisposable = extensions.onDidChange(() => {
      if (isOpenShiftConnectorInstalled()) {
        resolve();
      }
    });
    Promise.resolve(commands.executeCommand("workbench.extensions.installExtension", OPENSHIFT_CONNECTOR_EXTENSION_ID)).catch(reject);
    setTimeout(reject, DOWNLOAD_TIMEOUT, new Error(`${OPENSHIFT_CONNECTOR} installation is taking a while. Cancelling 'Deploy to OpenShift'. Please retry after the OpenShift Connector installation has finished`));
  }).finally(() => {
    installListenerDisposable.dispose();
  });
}
```

src/utils/openShiftConnectorUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { commands, Disposable, Extension, extensions, ProgressLocation, Uri, window } from "vscode";
+
+export const OPENSHIFT_CONNECTOR_EXTENSION_ID: string = 'redhat.vscode-openshift-connector';
+export const OPENSHIFT_CONNECTOR = 'OpenShift Connector extension';
+const DOWNLOAD_TIMEOUT: number = 60000; // Timeout for downloading VSCode OpenShift Connector, in milliseconds
```
datho7561: Remove typing

datho7561:
@fbricon This should work properly with the snapshot build of OpenShift Connector 0.2.7

dgolovin:
@datho7561 0.2.7 was released last week.

mohitsuman:
Hi @datho7561,  for this command, do we have any changes for it to be merged ?

datho7561:
@mohitsuman this PR should be good to merge. I'll merge it now.

PR #332, opened 2021-03-03
Fix dependencies

datho7561:
Remove force resolution for minimist, since none of the packages depend on the broken version any more.


datho7561:
Most of this PR has been added to other PRs

PR #338, opened 2021-03-16
Prompt to use TypeScript specified in package.json

datho7561:
Enables a prompt which alerts the vscode-quarkus developer when the version of the TypeScript language server used by VS Code is not the one specified in the `package.json`.


datho7561:
@fbricon

rgrunber (approved):
This fixes some of the issues I'm seeing with vscode vs. workspace typescript versions.

Issue #335, opened 2021-03-11
Study how to improve the native-image binary creation workflow


datho7561:
* How should the user get Mandrel?
* How should the user create a binary for local testing purposes?
* How should the user deploy a binary version of the app to OpenShift?

rgrunber:
>     * How should the user get Mandrel?

We could potentially download this on their behalf and place it accordingly. If the user already has it installed we can ask them to point to their installation.

Update : For clarification, this item should probably be lowest priority of the others (eg. we don't provide a JDK)

>     * How should the user create a binary for local testing purposes?

We have a task that handles launching a quarkus project in [dev mode](https://github.com/redhat-developer/vscode-quarkus/blob/master/src/wizards/debugging/TaskCreator.ts#L99), so we'll likely just need to take some of that code and modify the call to set the correct [profile/properties](https://github.com/quarkusio/quarkus/blob/master/devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java#L88) so the binary gets generated. I think this is probably the first thing to try to tackle if there's interest.

>     * How should the user deploy a binary version of the app to OpenShift?

Much like #313 I think redhat-developer/vscode-openshift-tools should be providing the API for this.

datho7561:
I am going to try and make a POC sets up the Maven binary compilation as a VS Code task in tasks.json ([see](https://code.visualstudio.com/docs/editor/tasks#vscode))

rgrunber:
As an update, we now have the ability to build a binary if `GRAALVM_HOME` or `JAVA_HOME` is set with an appropriate JVM. I think we should consider downloading Mandrel on behalf of the user (eg. https://github.com/graalvm/mandrel/releases ) as an option in https://github.com/redhat-developer/vscode-quarkus/blob/master/src/wizards/binary/buildBinary.ts#L45-L48 .

rgrunber:
Another possibility recently brought up by @angelozerr is the ability to run the binary on behalf of the user if we can detect it. When the command to build the binary is activated, it should be possible to learn the output location and use that to launch. Alternatively, maybe we could just search the entire project for the binary.

```
Produced artifacts:
 /home/rgrunber/sample-projects/my-quarkus-example/target/my-quarkus-example-1.0.0-SNAPSHOT-native-image-source-jar/my-quarkus-example-1.0.0-SNAPSHOT-runner (executable)
[...]

PR #341, opened 2021-03-18
Add a command to build a binary of a Quarkus app

datho7561:
* Adds a command that generates, then executes a `tasks.json` task that builds a binary
   * After you have used the command once, you can build a binary either through the command or `<Ctrl>+<Shift>+B` > Quarkus: Generate a binary
* The command checks to make sure GRAALVM_HOME is set before building the binary, and provides a link to the Quarkus docs if it's not set up.

This PR addresses one aspect of #335

### DIFF
--- a/USAGE_DATA.md
+++ b/USAGE_DATA.md
@@ -15,6 +15,7 @@ vscode-quarkus has opt-in telemetry collection, provided by [vscode-commons](htt
     * "Quarkus: Debug current Quarkus project"
     * "Quarkus: Generate a Quarkus project"
     * "Quarkus: Welcome"
+    * "Quarkus: Build executable"
 
 ## How to opt in or out
 
--- a/src/wizards/binary/buildBinary.ts
+++ b/src/wizards/binary/buildBinary.ts
@@ -1,0 +1,154 @@
+/**
+ * Copyright 2021 Red Hat, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as fs from "fs-extra";
+import * as path from "path";
+import { env, Task, tasks, Uri, window, workspace, WorkspaceFolder } from "vscode";
+import { BuildSupport } from "../../buildSupport/BuildSupport";
+import { ProjectLabelInfo } from "../../definitions/ProjectLabelInfo";
+import { TaskCreator } from "../debugging/TaskCreator";
+import { getQuarkusProject } from "../getQuarkusProject";
+import which = require("which");
+
+const NATIVE_IMAGE_DOCS_URI: Uri = Uri.parse('https://quarkus.io/guides/building-native-image');
+
+const RUN_ANYWAYS = 'Run anyways';
+const DO_NOT_RUN = 'Don\'t run';
+const LEARN_MORE = 'Learn more...';
+
+/**
+ * Selects a quarkus project, then starts building a binary for it
+ *
+ * Prompts the user if GraalVM isn't detected
+ *
+ * @returns when the binary build task has been started
+ */
+export async function buildBinary(): Promise<void> {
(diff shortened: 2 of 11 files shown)


src/wizards/debugging/TaskCreator.ts
```diff
@@ -87,52 +113,67 @@ export class TaskCreator {
...
+  /**
+   * Returns the task definition for building a native image of the Quarkus app
+   *
+   * @returns the task definition for building a native image of the Quarkus app
+   */
+  private async getQuarkusBinaryTask(): Promise<TaskDefinition> {
+    const taskLabel: string = this.quarkusBuildSupport.getQuarkusBinaryTaskName(this.workspaceFolder, this.projectFolder);
+    const unixCommand: string = (await this.quarkusBuildSupport.getQuarkusBinaryCommand(this.projectFolder, { windows: false })).command;
+    const windowsCommand: string = (await this.quarkusBuildSupport.getQuarkusBinaryCommand(this.projectFolder, { windows: true })).command;
+    const taskPatterns: TaskPattern = this.quarkusBuildSupport.getTaskPatterns();
```
datho7561: This specifies to use the same TaskPatterns as for quarkus:dev, but the build tool output is different between quarkus:dev and the binary build

datho7561:
~~Demo: https://youtu.be/5WgSg_FzzN8~~ Sorry, didn't realize that I put the wrong clip in the demo video

New Demo: https://youtu.be/qleIxqrqbLs

datho7561:
I rebased this PR and added a section about collecting telemetry when the command is run.

USAGE_DATA.md
```diff
@@ -15,6 +15,7 @@ vscode-quarkus has opt-in telemetry collection, provided by [vscode-commons](htt
     * "Quarkus: Debug current Quarkus project"
     * "Quarkus: Generate a Quarkus project"
     * "Quarkus: Welcome"
+    * "Quarkus: Build a binary"
```
rgrunber: Thoughts on "Build a binary" vs. "Build executable". Or "Generate Executable"

src/wizards/binary/buildBinary.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+/**
+ * Returns true if GRAALVM_HOME is set and false otherwise
+ *
+ * @returns true if GRAALVM_HOME is set and false otherwise
+ */
+function hasGraalVmHome(): boolean {
+  return !!process.env.GRAALVM_HOME.trim();
```
rgrunber: Running without GRAALVM_HOME set : `Exception has occurred: TypeError: Cannot read property 'trim' of undefined`

src/wizards/binary/buildBinary.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @returns when the binary build task has been started
+ */
+export async function buildBinary(): Promise<void> {
+  const projectLabelInfo: ProjectLabelInfo = await getQuarkusProject();
+  if (!projectLabelInfo) {
+    return;
+  }
+  if (!hasGraalVmHome()) {
+    const result = await window.showInformationMessage(
+      'The environment variable GRAALVM_HOME is not set, '
```
rgrunber: If we wanted to confirm whether JAVA_HOME or GRAALVM_HOME contains a GraalVM/Mandrel JDK, could we do that ? Also PATH could also contain the bin path of the Graal/Mandrel JDK and that seems to work.

I just mention these because they're valid options for a user following the configuration in the linked configuration web page.

rgrunber (commented):
Seems to be working pretty well. Just ran into a small issue, and maybe we could have better support for other environment variables.

USAGE_DATA.md
datho7561: I think I like "Build executable" better, so I'll change it to that

src/wizards/binary/buildBinary.ts
datho7561: We could search `$GRAALVM_HOME/bin` for `native-image` to identify a Java runtime that can build native images, and we can search `$GRAALVM_HOME/bin` for `node` to distinguish GraalVM from Mandrel (GraalVM ships with node)

datho7561: ^ and do the above with JAVA_HOME if GRAALVM_HOME is not set, or as a last resort, do [which](https://www.npmjs.com/package/which) `native-image`. I'll work on this.

rgrunber (approved):
I think just updating the README with the new command.

src/wizards/debugging/TaskCreator.ts
datho7561: Removed task patterns for build executable

rgrunber (approved):
Change looks good. I think it should be fine to merge.

PR #345, opened 2021-03-25
Emit telemetry events when commands are executed

### LINKED ISSUE #344: Emit telemetry events for command executions
It would be helpful to collect telemetry on which of the commands provided by vscode-quarkus are used. Emit telemetry events when each of the commands are run successfully. Document that these telemetry events are emitted.

datho7561:
Please refer to USAGE_DATA.md to see what telemetry is collected. Emits telemetry events for command run failure/success, and attaches errors for command failures.

Closes #344

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -1,0 +1,108 @@
+import { commands, ExtensionContext, window } from "vscode";
+import { VSCodeCommands } from "../definitions/constants";
+import { ProjectLabelInfo } from "../definitions/ProjectLabelInfo";
+import { requestStandardMode } from "../utils/requestStandardMode";
+import { sendCommandFailedTelemetry, sendCommandSucceededTelemetry } from "../utils/telemetryUtils";
+import { WelcomeWebview } from "../webviews/WelcomeWebview";
+import { addExtensionsWizard } from "../wizards/addExtensions/addExtensionsWizard";
+import { startDebugging } from "../wizards/debugging/startDebugging";
+import { generateProjectWizard } from "../wizards/generateProject/generationWizard";
+
+const NOT_A_QUARKUS_PROJECT = new Error('No Quarkus projects were detected in this folder');
+const STANDARD_MODE_REQUEST_FAILED = new Error('Error occurred while requesting standard mode from the Java language server');
+
+/**
+ * Register the vscode-quarkus commands
+ *
+ * @param context the extension context
+ */
+export function registerVSCodeCommands(context: ExtensionContext): void {
+
+  /**
+   * Command for creating a Quarkus project
+   */
+  registerCommandWithTelemetry(context, VSCodeCommands.CREATE_PROJECT, generateProjectWizard);
+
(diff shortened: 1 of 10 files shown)


fbricon:
it's valuable to understand when commands failed to execute, so we should send the execution outcome (successful or not), with the error message in case it failed

datho7561:
Right now, the only cases where the event doesn't get sent is if the command requires a Quarkus project to run on and none is present. I'll have to restructure some of the code in order to capture any failure messages

datho7561:
Some of the error messages might contain PII, which is going to complicate this. I can catch and rethrow the errors with generic messages. In order for the messages to be useful to us, I would need to do this at a fine-grained level.

fbricon:
@datho7561 I'm working on automatically sanitizing all event properties (in vsode-redhat-telemetry), ie hiding paths and all references to the user name, so we should be fine

rgrunber:
Just wondering. The events are ${command}_success and ${command}_failure . For the sake of collection would it not make more sense to call the event ${command}, and set the status to success/failure in its properties ? This would be similar to how we do it for vscode-xml.

src/commands/registerCommands.ts
```diff
@@ -0,0 +1,104 @@
...
+ * @param context the extension context
+ * @param commandName the name of the command to register
+ * @param commandAction the async function to run when the command is called
+ */
+async function registerCommandWithTelemetry(context: ExtensionContext, commandName: string, commandAction: () => Promise<any>): Promise<void> {
+  context.subscriptions.push(commands.registerCommand(commandName, async () => {
+    try {
+      await commandAction();
+      sendCommandSucceededTelemetry(commandName);
+    } catch (e) {
```
rgrunber, line 58: `e` could be both a String or an error our code, right ? The arguments of the 2 methods below expect e to be a string but afaict it's most likely going to be an Error, so seems like the type should be tested, or `e.msg` should be used.

datho7561:
> ...call the event ${command}, and set the status to success/failure in its properties ?...

Should be good now

src/commands/registerCommands.ts
datho7561: I tested the type to get the correct error message. I don't know if any of the libraries we use throw strings instead of errors when there is an error.

rgrunber: Yeah I noticed we throw some Strings in our code but mostly I could see Errors for sure. I was just throwing an Error right before calling a command to try and see if it got picked up in telemetry, but it was coming back empty. This should fix it.

rgrunber:
Works for me. Another thing to consider though. In vscode-xml we (eventually) went with `.` as a separator for event names. We should probably do the same here. (ie. not convert them to `_`) .

datho7561:
I removed the code that replaces the periods with underscores

rgrunber (approved):
Change looks good to me.

I think maybe in the future we can consider expanding the addextension to also report which extensions, but this is a good first step.

PR #369, opened 2021-06-23
Update @redhat-developer/vscode-redhat-telemetry to 0.2.0

### LINKED ISSUE #365: Update @redhat-developer/vscode-redhat-telemetry to 0.1.1
This removes the dependency to vscode-commons.

See https://github.com/redhat-developer/vscode-redhat-telemetry#add-the-below-code-to-your-extensionts for instructions on how to implement this change.

~~Note that `redhat.telemetry.enabled` doesn't need to be added to the `package.json`, since it will be contributed by vscode-java~~ see note from Fred

datho7561:
Closes #365

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,7 @@
   "extensionDependencies": [
     "redhat.vscode-microprofile",
     "redhat.java",
-    "vscjava.vscode-java-debug",
-    "redhat.vscode-commons"
+    "vscjava.vscode-java-debug"
   ],
   "contributes": {
     "javaExtensions": [
@@ -292,7 +291,7 @@
     "webpack-cli": "^3.3.2"
   },
   "dependencies": {
-    "@redhat-developer/vscode-redhat-telemetry": "0.0.18",
+    "@redhat-developer/vscode-redhat-telemetry": "^0.2.0",
     "ejs": "^2.7.1",
     "expand-home-dir": "0.0.3",
     "find-up": "^4.1.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -147,6 +146,12 @@
     "configuration": {
       "title": "Quarkus Tools",
       "properties": {
+        "redhat.telemetry.enabled": {
```
rgrunber: According to https://github.com/redhat-developer/vscode-redhat-telemetry#contribute-the-redhattelemetryenabled-preference , `Unless your extension already depends on a telemetry-enabled Red Hat extension, it needs to declare the redhat.telemetry.enabled preference in its package.json, like:` . So does this need to be here ? This depends on vscode-microprofile, which defines this property.

rgrunber (changes requested):
I do seem to be getting the warning about vscode-quarkus being unable to register redhat.telemetry.enabled because it's already registered.

Otherwise, startup/shutdown/commands seem to be reported correctly.

fbricon:
No, redhat.telemetry.enabled shouldn't be defined here, as per the doc

Le ven. 23 juil. 2021 à 01:34, Roland Grunberg ***@***.***> a
écrit :

src/utils/telemetryUtils.ts
```diff
@@ -30,10 +44,14 @@ export async function sendCommandFailedTelemetry(commandName: string, msg?: stri
  *
  * @param commandName the name of the command that was run
  * @param suffix the suffix to add to the command to get the event name
+ * @throws if the telemetry service has not been initialized yet
+ * @returns when the telemetry
```
rgrunber: "when the telemetry event has been sent"

rgrunber (approved):
Change seems to be working for me. Tested on a vsix.

datho7561:
Thanks for the review, Roland!

PR #370, opened 2021-06-28
Consume vscode-microprofile API

### LINKED ISSUE #323: Investigate running Quarkus project generator without waiting for Java extension

datho7561:
See redhat-developer/vscode-microprofile#58

Consume the API provided by vscode-microprofile in order to wait for lsp4mp to start when running commands that interact with lsp4mp.

Closes #323

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,8 +51,6 @@
   "main": "./dist/extension",
   "extensionDependencies": [
     "redhat.vscode-microprofile",
-    "redhat.java",
-    "vscjava.vscode-java-debug",
     "redhat.vscode-commons"
   ],
   "contributes": {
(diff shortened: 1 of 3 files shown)


datho7561:
@fbricon I implemented something close to the approach you suggested. vscode-microprofile would no longer depend on vscode-java, and would handle prompting to install it. vscode-microprofile exits its `activate()` method before lsp4mp is started, then has an API that lets you check once its started. vscode-quarkus still depends on vscode-microprofile, and uses the API to tell when it can use features of lsp4mp (detecting which projects are microprofile projects). This means a user can access the project generation wizard from rootless mode. However, vscode-java and jdt-ls still end up starting, since starting vscode-microprofile means starting lsp4mp, which means starting jdt.ls.

datho7561:
See comments on https://github.com/redhat-developer/vscode-microprofile/pull/58

Issue #371, opened 2021-07-07
Add support for configuration file per profile


fbricon:
> The user can have an application.properties with general configuration and one file for each profile, like application-dev.profile for dev profile.

- https://github.com/quarkusio/quarkus/pull/11145

> Properties in the profile aware file have priority over profile aware properties defined in the main file.

- https://github.com/quarkusio/quarkus/issues/18292

datho7561:
I've worked on eclipse/lsp4mp#149, which should make it easier to implement this feature. The PR also lets us implement the support for per profile config files in quarkus-ls, so that the base lsp4mp project doesn't need to support it.

PR #373, opened 2021-07-07
Recognize Quarkus per profile config files

datho7561:
`application-*.properties` are now recognized as Quarkus Properties files, and will benefit from autocompletion of property keys and values.

See the language server side changes here: https://github.com/redhat-developer/quarkus-ls/pull/411

One portion of #371

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "onCommand:quarkusTools.createProject",
     "onCommand:quarkusTools.welcome",
     "workspaceContains:**/src/main/resources/application.properties",
+    "workspaceContains:**/src/main/resources/application-?*.properties",
     "workspaceContains:**src/main/resources/META-INF/microprofile-config.properties",
     "workspaceContains:**/src/main/resources/application.yaml",
     "workspaceContains:**/src/main/resources/application.yml",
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -124,6 +124,8 @@ function tryToForceLanguageId(document: TextDocument, fileName: string, document
   });
 }
 
+const APP_PROPERTIES_PATTERN = /^application(?:-[^.]+)\.properties$/;
```
datho7561: On the quarkus-ls side, I used `[A-Za-z]` to match the profile name; figure out if non letter characters are supported by Quarkus.

package.json
```diff
@@ -72,12 +72,13 @@
     "languages": [
       {
         "id": "quarkus-properties",
+        "filenamePatterns": [
+          "application.properties",
+          "application-?*.properties"
```
rgrunber, line 78: This ensures the language id is set correctly. However, I noticed that the extension doesn't start until the file is opened, while with just application.properties it does.

Is it in the scope of this change to add it to activationEvents or does that not support globs in this way.

datho7561: It's definitely in scope. I'll take look into it

datho7561: Okay, should be fixed

Issue #322, opened 2020-11-20
'RESTEasy JAX-RS' should no longer be added by default


fbricon:
https://code.quarkus.io/ no longer adds 'RESTEasy JAX-RS' by default, so the new project wizard should do the same

fbricon:
It's not that trivial since https://code.quarkus.redhat.com/api is behind https://code.quarkus.io/api, and still includes it by default.

@ia3andy any way to detect whether we should flag 'RESTEasy JAX-RS' as included?

ia3andy:
@fbricon yes, on https://code.quarkus.redhat.com/api/extensions it has the `default` flag to true, and to false on https://code.quarkus.io/api/extensions

datho7561:
Sorry, just looking at this now. If I use the default options through the web UI at [code.quarkus.io](https://code.quarkus.io), or specify 0 extensions when querying the rest API, I still get this in the pom.xml:

```
...
<dependency>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-resteasy</artifactId>
</dependency>
...
```

Is RESTEasy JAX-RS separate from quarkus-resteasy? Is RESTEasy JAX-RS is automatically added unless some other component is added? If so, which component do I need to add to prevent RESTEasy JAX-RS from being automatically added?

ia3andy:
@datho7561 if you don't pick any other example, then the `quarkus-resteasy` example is added.

If you don't want any example, you need to add an extension with anouther example (ex: `quarkus-qute`) or to add `&ne=trye` which stand for `no examples`.

datho7561:
@fbricon We could add another step to the wizard to specify if quickstart code should be generated, and add a warning that adding quickstart code may pull in other dependencies. Do you think this would be a good solution?

ia3andy:
Hi @datho7561 ,

The api has changed since the discussion started, the parameter to use is `nc` ("No Code"), `ne` has been deprecated, the text to use in the wizard would be:
"Decide whether to include the starter code provided by extensions or get an empty Quarkus project."
If the answer is "Yes" then `nc=false`
If the answer is "No" then `nc=true`

It  would be nice to update this before it gets to delivered to the user.


The reason behind this is that we don't deliver sample code anymore but starter code, the nuance is subtile (starter code is meant to start your app while sample/example code is meant to learn and be deleted)

Let me know if you have any question :)

Cheers

fbricon:
[...]

PR #382, opened 2021-07-28
Use `nc` parameter if available when generating project

datho7561:
Checks the Code Quarkus openapi for the `nc` parameter. If it is available for the given Code Quarkus API, then it will pass `nc=true` to the project generation API when the user wants to exclude example code, instead of the deprecated `ne=true` parameter.

Addresses comments on #322

### DIFF
--- a/src/definitions/inputState.ts
+++ b/src/definitions/inputState.ts
@@ -34,7 +34,7 @@ export interface ProjectGenState extends State {
   packageName: string;
   resourceName: string;
   targetDir: Uri;
-  isGenerateSampleCode: boolean;
+  shouldGenerateCode: boolean;
 }
 
 export interface AddExtensionsState extends State {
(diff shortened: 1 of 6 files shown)


src/definitions/inputState.ts
```diff
@@ -34,7 +34,7 @@ export interface ProjectGenState extends State {
   packageName: string;
   resourceName: string;
   targetDir: Uri;
-  isGenerateSampleCode: boolean;
+  isGenerateCode: boolean;
```
rgrunber: Might this make more sense as `shouldGenerateCode` ?

datho7561:
Tests fail locally; I am curious to see if they pass on GH Actions

rgrunber (approved):
I've tried this out and seems to work for me.

PR #384, opened 2021-08-04
Increase test timeouts and fix OpenAPI download

### LINKED ISSUE #379: Tests time out on macOS on GitHub Actions
Sometimes the UI tests fail at the 'remove temp dir' step. This is a cleanup step that runs after all the tests are done. The reason of the failure is that this step sometimes times out. Investigate different methods of resolving this test failure.

datho7561:
The code to handle downloading the OpenAPI YAML file had a bug which caused it to intermittently fail. As the size of the YAML document got larger, the failures became more consistent, so I was able to pinpoint the issue.

Closes #379

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,10 +5,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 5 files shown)


rgrunber (approved):
LGTM. I'm not able to reproduce the failure locally with test-ui but the change makes sense.

datho7561:
Thanks for the review, Roland!

PR #388, opened 2021-08-24
Add launch configuration for also debugging MicroProfile extension

rgrunber:
- It is sometimes necessary to debug both the Quarkus and MicroProfile
  project folders at once
- extensionDevelopmentPath option may be provided multiple times,
  permitting the debugging of multiple extensions at once
- The new configuration assumes a contributor has placed the
  vscode-quarkus & vscode-microprofile project folders under the same
  folder

See https://github.com/microsoft/vscode/issues/72500 . Though not heavily documented, this seems to be working for me, and not having to build .vsix or do some .vscode/extension symlinking seems very convenient.


rgrunber:
Let's add this configuration to the CONTRIBUTING file : https://github.com/redhat-developer/vscode-quarkus/blob/master/CONTRIBUTING.md#testing-vscode-quarkus

.vscode/launch.json
```diff
@@ -21,6 +21,24 @@
       ],
       "preLaunchTask": "npm: watch"
     },
+    {
+      "name": "Run Extension & vscode-microprofile",
```
rgrunber: I'm also open to changing the name :

- "Run Extension with MicroProfile"
- "Run Quarkus & MicroProfile Extensions"
...

angelozerr: What about `Run Extension with installed MicroProfile`?

angelozerr: and the second like this (which is not a perfect english) `Run Extension with sources MicroProfile`

rgrunber: I would think "installed" implies from the marketplace. I think "Run Extension with Development MicroProfile" or "Run Extension with MicroProfile Project" gets closer to what I'd like.

rgrunber: I think "Run Extension with MicroProfile sources" could work.

angelozerr: Installed means to install from vsix too, no?

rgrunber: Yes. that as well. That's why I prefer the term "source" or "development". It makes it clear we're running the project folder and that no .vsix or marketplace content is consumed.

angelozerr:
I'm Ok to merge the PR. @rgrunber do you want to update something?

rgrunber:
> I'm Ok to merge the PR. @rgrunber do you want to update something?

No, I think this is fine. Has the UI of VS Code changed since the guide was done ? I noticed the old entries had the extension mentioned in round brackets, and even things like terminal icon were present. My image doesn't match exactly, but it's close enough :P

PR #390, opened 2021-08-30
Support termination of projects with no starter code.

### LINKED ISSUE #389: Debugging no-starter code projects fails
When creating a no-starter code project (you still need to create 1 java file), if you execute the debug quarkus command, the process never completes.
This is mostly caused by the quarkus dev task's problemMatcher.background.endspattern value, that expects a "listening on" pattern. 
https://github.com/redhat-developer/vscode-quarkus/blob/46b180ee55cbf71ccef64e04c90bdc273f3b7146/src/buildSupport/GradleBuildSupport.ts#L30

https://github.com/redhat-developer/vscode-quarkus/blob/46b180ee55cbf71ccef64e04c90bdc273f3b7146/src/buildSupport/MavenBuildSupport.ts#L30

When no http server is started, that last sentence is missing, it then just waits indefinitely.

Using `^.*Quarkus .* started in .*\\.*` works.

Also, it doesn't seem to cause an issue but when a Gradle deamon is reused, this pattern is never found:
https://github.com/redhat-developer/vscode-quarkus/blob/46b180ee55cbf71ccef64e04c90bdc273f3b7146/src/buildSupport/GradleBuildSupport.ts#L29

I'd simply use `Task :` instead

rgrunber:
- Adjust task pattern to recognize build output when no starter code is
  present
- Fixes #389

### DIFF
--- a/src/buildSupport/GradleBuildSupport.ts
+++ b/src/buildSupport/GradleBuildSupport.ts
@@ -26,8 +26,8 @@ export class GradleBuildSupport extends BuildSupport {
       quarkusBinary: 'buildNative',
       wrapper: 'gradlew',
       wrapperWindows: 'gradlew.bat',
-      taskBeginsPattern: '^.*Starting a Gradle Daemon*',
-      taskEndsPattern: '^.*Quarkus .* started in .*\\. Listening on:*'
+      taskBeginsPattern: '^.*Task: ',
+      taskEndsPattern: '(^.*Quarkus .* started in .*\\.)|(^.* ERROR .* Failed to start)'
     });
   }
 
--- a/src/buildSupport/MavenBuildSupport.ts
+++ b/src/buildSupport/MavenBuildSupport.ts
@@ -26,8 +26,8 @@ export class MavenBuildSupport extends BuildSupport {
       quarkusBinary: 'package -Pnative',
       wrapper: 'mvnw',
       wrapperWindows: 'mvnw.cmd',
-      taskBeginsPattern: '^.*Scanning for projects...*',
-      taskEndsPattern: '^.*Quarkus .* started in .*\\. Listening on:*'
+      taskBeginsPattern: '^.*Scanning for projects...',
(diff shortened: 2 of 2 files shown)


rgrunber:
I'm pretty sure the '*' at the very end aren't needed, since they should just mean 0 or more of the previous term. Either way, this seems to work for me.

rgrunber:
Let me see if I can catch failed launches in the end pattern. No point in having the task show continuous loading on failures.

Also, for information : We use `activeOnStart: true` : `If set to true the background monitor is in active mode when the task starts. This is equals of issuing a line that matches the beginsPattern`. So I guess the begin pattern doesn't make a difference. Given how the beginning line is pretty much guaranteed, I don't necessarily think it's a bad thing.

src/buildSupport/GradleBuildSupport.ts
```diff
@@ -26,8 +26,8 @@ export class GradleBuildSupport extends BuildSupport {
       quarkusBinary: 'buildNative',
       wrapper: 'gradlew',
       wrapperWindows: 'gradlew.bat',
-      taskBeginsPattern: '^.*Starting a Gradle Daemon*',
-      taskEndsPattern: '^.*Quarkus .* started in .*\\. Listening on:*'
+      taskBeginsPattern: '^.*Task: ',
+      taskEndsPattern: '^.*(Quarkus .* started in .*\\.| ERROR .* Failed to start)'
```
fbricon: don't you need a dollar sign to capture the end of the group? It's unclear this matches (success)|(failure) patterns

rgrunber: It matches the existence of either pattern and doesn't have to match the full line (by including a dollar sign). I just figured there's no point in saying ".*$" when simply omitting it implies the same thing.

rgrunber: Done. Round brackets are now of form (success)|(failure) and that seems to work just fine.

fbricon (approved):

PR #394, opened 2021-09-17
added Quarkus platform version selection to project generation wizard

### LINKED ISSUE #393: Support selecting Quarkus platform version in project wizard
There is now an option to select the Quarkus platform version on https://code.quarkus.io/ . The query string parameter controlling this appears to be `S`. See https://github.com/redhat-developer/vscode-quarkus/blob/master/src/wizards/generateProject/generationWizard.ts & https://github.com/redhat-developer/vscode-quarkus/blob/master/src/utils/requestUtils.ts#L69

AlexXuChen:
Added Quarkus platform version selection from `io.quarkus.platform` as step in project generation wizard.

Closes #393 


AlexXuChen:
Currently the input box validation isn't working, it defaults to the `groupId` validation.

AlexXuChen:
Also need retrieval from `download?`

AlexXuChen:
How to test:

Platforms 2.2 and 2.1 differ by the extension "Nayanara LRA".

if you select `2.2 (recommended)`, then in the "Extension Picker", you should see the following when searching "Nayanara LRA":
Screenshot from 2021-09-20 17-28-01

if you select `2.1`, then in the "Extension Picker", you should see the following when searching "Nayanara LRA":
Screenshot from 2021-09-20 17-28-18

AlexXuChen:
Currently fixing issue where Quarkus project won't download

src/definitions/constants.ts
```diff
@@ -39,6 +39,7 @@ export const VSCODE_QUARKUS_EXTENSION_NAME = "redhat.vscode-quarkus";
 
 export const DEFAULT_API_URL: string = 'https://code.quarkus.io/api';
 export const DEFAULT_BUILD_TOOL: BuildToolName = BuildToolName.Maven;
+export const DEFAULT_PLATFORM_VERSION: string = 'io.quarkus.platform:2.2';
```
rgrunber: I don't see this constant being used anywhere. In fact when computing the the versions to populate the dropdown, you use the "recommended" value from the API request, so it looks like there's no need for this.

I see this as a good thing because this value would have had to be updated everytime a new version were released anyways.

src/utils/codeQuarkusApiUtils.ts
```diff
@@ -38,6 +39,15 @@ export interface CodeQuarkusFunctionality {
   supportsNoCodeParameter: boolean;
 }
 
+/**
+ * Represents the the response object from the `/streams` rout from the Code Quarkus API
```
rgrunber: Typo at "rout" but I would use "endpoint" instead.

src/wizards/generateProject/generationWizard.ts
```diff
@@ -58,14 +58,35 @@ export async function generateProjectWizard() {
...
+    // Place the recommended option first in the quick pick list
+    quickPickItems.sort((x: PlatformVersionPickItem, y: PlatformVersionPickItem) => {
+      return (x.recommended === y.recommended) ? 0 : x.recommended ? -1 : 1;
+    });
+
+    state.platformVersion = (await input.showQuickPick({
+      title: INPUT_TITLE,
+      step: input.getStepNumber(),
+      totalSteps: state.totalSteps,
+      placeholder: 'Pick io.quarkus.platform',
```
rgrunber: Minor nitpicks :

- "Pick io.quarkus.platform version"
- If we're going to use "io.quarkus.platform" in the instruction/hint, then there's no need to include it also in each item. I would just have each item label show the version (eg. 2.2, 2.1) and a label like "2.2 (recommended)" where recommended is true.

src/wizards/generateProject/generationWizard.ts
```diff
@@ -58,14 +58,35 @@ export async function generateProjectWizard() {
...
+    return (input: MultiStepInput) => inputPlatformVersion(input, state);
+  }
+
+  async function inputPlatformVersion(input: MultiStepInput, state: Partial<ProjectGenState>) {
+
+    const quickPickItems: PlatformVersionPickItem[] = await getCodeQuarkusApiPlatforms();
+
+    // Place the recommended option first in the quick pick list
+    quickPickItems.sort((x: PlatformVersionPickItem, y: PlatformVersionPickItem) => {
+      return (x.recommended === y.recommended) ? 0 : x.recommended ? -1 : 1;
```
rgrunber: Should we sort by `quarkusCoreVersion` when both have the same recommendation value ? Not sure if the `/streams` response maintains any ordering.

src/QuarkusContext.ts
```diff
@@ -67,6 +68,7 @@ export namespace QuarkusContext {
   export function setDefaults(defaults: Defaults): void {
     checkContext();
     context.globalState.update(BUILD_TOOL, defaults.buildTool);
+    context.globalState.update(PLATFORM_VERSION, defaults.platformVersion);
```
rgrunber: The whole point of this function is to save the values that were used to the globalState database so the next time someone uses the wizard, we set the default values to what were used the previous time.

You're saving the last used platform version as the default, but you never retrieve it. The default just ends up being the recommended one. I'm fine with that, but then there's no point in having this line or the one about PLATFORM_VERSION.

rgrunber (commented):
Looks good so far.

src/wizards/generateProject/generationWizard.ts
AlexXuChen: Would it be possible for multiple platforms to be recommended?

rgrunber: I think it could happen if there were many choices for versions, but for a small amount it would be a bit strange. With that said, we should still probably assume it may happen.

src/QuarkusContext.ts
AlexXuChen: Should we use the default from https://github.com/redhat-developer/vscode-quarkus/pull/394#discussion_r713196041?

rgrunber: No. I wouldn't use a default.

src/wizards/generateProject/generationWizard.ts
```diff
@@ -58,14 +58,36 @@ export async function generateProjectWizard() {
...
+    return (input: MultiStepInput) => inputPlatformVersion(input, state);
+  }
+
+  async function inputPlatformVersion(input: MultiStepInput, state: Partial<ProjectGenState>) {
+
+    const quickPickItems: PlatformVersionPickItem[] = await getCodeQuarkusApiPlatforms();
+
+    // Place the recommended option first in the quick pick list
+    quickPickItems.sort((x: PlatformVersionPickItem, y: PlatformVersionPickItem) => {
+      return (x.recommended === y.recommended) ? (x.quarkusCoreVersion < y.quarkusCoreVersion ? -1 : 1) : x.recommended ? -1 : 1;
```
rgrunber: quarkusCoreVersion is a string so this comparison won't work how you think. https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUAiAjABnr76ABADyHoBMee6AlEA

I'd also just update the comment above to mention that we now sort by recommended, and then use version..

src/wizards/generateProject/generationWizard.ts
```diff
@@ -58,14 +58,68 @@ export async function generateProjectWizard() {
...
+
+  async function inputPlatformVersion(input: MultiStepInput, state: Partial<ProjectGenState>) {
+
+    const quickPickItems: PlatformVersionPickItem[] = await getCodeQuarkusApiPlatforms();
+
+    // Sort by recommended and quarkusCoreVersion in case of tie in the quick pick list
+    quickPickItems.sort((x: PlatformVersionPickItem, y: PlatformVersionPickItem) => {
+      return (x.recommended === y.recommended) ? (quarkusCoreVersionComparator(x.quarkusCoreVersion.split("."), y.quarkusCoreVersion.split(".")) ? -1 : 1) : x.recommended ? -1 : 1;
+
+      function quarkusCoreVersionComparator(x: string[], y: string[]) {
```
AlexXuChen: This should cover any length of quarkusCoreVersion comparison

*UPDATE - they just added version 2.3, where 2.1 and 2.3 are both not recommended, and even though 2.3 appears first in the response, it is sorted properly (after 2.1)

The response object: 
```json
[
    {
        "key": "io.quarkus.platform:2.3",
        "quarkusCoreVersion": "2.3.0.CR1",
        "recommended": false
    },
    {
        "key": "io.quarkus.platform:2.1",
        "quarkusCoreVersion": "2.1.4.Final",
        "recommended": false
    },
    {
        "key": "io.quarkus.platform:2.2",
        "quarkusCoreVersion": "2.2.3.Final",
        "recommended": true
    }
]
```
The wizard:
Screenshot from 2021-09-22 12-04-01

src/wizards/generateProject/generationWizard.ts
AlexXuChen: https://github.com/redhat-developer/vscode-quarkus/pull/394#discussion_r714089771

src/wizards/generateProject/generationWizard.ts
AlexXuChen: Number.isNaN works on my machine, but has an error thrown on CI

AlexXuChen:
They just added version 2.3, the wizard should be sorted in descending order, with recommended first

The response object: 
```json
[
    {
        "key": "io.quarkus.platform:2.3",
        "quarkusCoreVersion": "2.3.0.CR1",
        "recommended": false
    },
    {
        "key": "io.quarkus.platform:2.1",
        "quarkusCoreVersion": "2.1.4.Final",
        "recommended": false
    },
    {
        "key": "io.quarkus.platform:2.2",
        "quarkusCoreVersion": "2.2.3.Final",
        "recommended": true
    }
]
```
The wizard:
Screenshot from 2021-09-22 15-05-23

src/wizards/generateProject/generationWizard.ts
rgrunber: I would avoid writing such things from scratch. There is almost always a library that can do it. For example https://www.npmjs.com/package/semver . I think to avoid adding a new library just for this, you could take the item's key, and use the string to the right of ':', which seems to be a float (eg. 2.2) as opposed to a version (eg. 2.2.3).

I think parseFloat should be able to convert it.

AlexXuChen: This would no longer sort by quarkusCoreVersion, since the quarkusCoreVersion typically comes as x.x.x, correct?

src/wizards/generateProject/generationWizard.ts
```diff
@@ -58,14 +58,36 @@ export async function generateProjectWizard() {
...
+    return (input: MultiStepInput) => inputPlatformVersion(input, state);
+  }
+
+  async function inputPlatformVersion(input: MultiStepInput, state: Partial<ProjectGenState>) {
+
+    const quickPickItems: PlatformVersionPickItem[] = await getCodeQuarkusApiPlatforms();
+
+    // Sort by recommended and version number in case of tie in the quick pick list
+    quickPickItems.sort((x: PlatformVersionPickItem, y: PlatformVersionPickItem) => {
+      return (x.recommended === y.recommended) ? (parseFloat(x.label) > parseFloat(y.label) ? -1 : 1) : x.recommended ? -1 : 1;
```
rgrunber, line 78: I wouldn't use x.label, because it'll be something like "2.2 (recommended)" where recommended is true. You can just redefine `x["key"].split(":")[1]` and use the result of that.

src/wizards/generateProject/generationWizard.ts
```diff
@@ -228,6 +250,7 @@ function saveDefaults(state: ProjectGenState): void {
 
   QuarkusContext.setDefaults({
     buildTool: state.buildTool,
+    platformVersion: state.platformVersion,
```
rgrunber: I think this can be removed. We're not currently using it.

src/QuarkusContext.ts
```diff
@@ -84,6 +84,7 @@ export namespace QuarkusContext {
 
 interface Defaults {
   buildTool: string;
+  platformVersion: string;
```
rgrunber: This can also be removed since we're not making use of the default value for now.

src/wizards/generateProject/generationWizard.ts
AlexXuChen: parseFloat("2.2 (recommended)") will give 2.2: https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUAHAhgJwrAYkpgC6oBEATPBQASraygC2TsYAJrO2QJQ9A

rgrunber: Ok, let's leave it as is. Turns out this is well described behaviour https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat#examples .

PR #407, opened 2021-10-21
ignored resource name step in wizard if no starter code included

### LINKED ISSUE #396: When generating a project without starter code, the Resource name step should be skipped
Right now, when generating a quarkus project, you set the Resources name and at the end, decide you don't want to include starter code. This is backward. You should select if starter code is included and then, if it's still relevant, ask to about the resource name.

AlexXuChen:
Ignored resource name step in project generation wizard if no starter code included.

Fixes #396 

### DIFF
--- a/src/wizards/generateProject/ExtensionsPicker.ts
+++ b/src/wizards/generateProject/ExtensionsPicker.ts
@@ -112,7 +112,7 @@ export class ExtensionsPicker {
       pick = await input.showQuickPick<QuickPickExtensionItem, QuickPickParameters<QuickPickExtensionItem>>({
         title: 'Quarkus Tools',
         step: this.options.step ? this.options.step : input.getStepNumber(),
-        totalSteps: state.totalSteps,
+        totalSteps: state.shouldGenerateCode ? state.totalSteps : state.totalSteps - 1,
         placeholder: 'Pick extensions',
         items: quickPickItems,
         buttons: [ this.getInfoButton() ],
--- a/src/definitions/inputState.ts
+++ b/src/definitions/inputState.ts
@@ -22,6 +22,7 @@ export interface State {
   totalSteps: number;
   extensions: QExtension[];
   platformVersion: string;
+  shouldGenerateCode: boolean;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


src/wizards/generateProject/ExtensionsPicker.ts
```diff
@@ -112,7 +112,7 @@ export class ExtensionsPicker {
       pick = await input.showQuickPick<QuickPickExtensionItem, QuickPickParameters<QuickPickExtensionItem>>({
         title: 'Quarkus Tools',
         step: this.options.step ? this.options.step : input.getStepNumber(),
-        totalSteps: state.totalSteps,
+        totalSteps: state.shouldGenerateCode ? state.totalSteps : state.totalSteps - 1,
```
AlexXuChen, line 115: Hacky implementation: I change the total steps sent to the input instead of the actual total steps in the state

AlexXuChen: @fbricon Thoughts on this fix?

fbricon: So I tested it and it works. What I don't grok is how the totalSteps is reset once you go back to change the shouldGenerateCode value in the previous page.

fbricon: nervermind, I turned on my brains, I got it now. Yes it's hacky. I can live with that.

AlexXuChen: The totalSteps displayed in the extension picker is updated based on shouldGenerateCode, instead of updating the state totalsSteps. This way, if you press the back button, any step that accesses state.totalSteps will still see 9 instead of 8, and if shouldGenerateCode is true, you'll see 8/9 in the ResourceName step instead of 8/8 in the ExtensionPicker step.

rgrunber (approved):
Works for me.

rgrunber:
Oh, and in case anyone noticed the failures on `npm run watch`, it's because webpack-cli eliminated the `--info-verbosity verbose` flag in webpack-cli 4.x and I updated to that in https://github.com/redhat-developer/vscode-quarkus/pull/410 . We may just have to also remove the patterns for it in `.vscode/tasks.json` since without the flag I'm not sure we can determine proper success of the command.

PR #409, opened 2021-10-25
Make "Deploy to OpenShift" command more discoverable

### LINKED ISSUE #406: Make the "Deploy to OpenShift" command more discoverable
Today, only people who know the "Deploy to OpenShift"  command exist will be able to use it. We need to figure out a way to make it more obvious

rgrunber:
- Fixes #406

deploy-quarkus-editor-titlebar-menu

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,10 @@
         "title": "Quarkus: Debug current Quarkus project"
       },
       {
+        "command": "quarkusTools.debugQuarkusProject.short",
+        "title": "Debug Quarkus"
+      },
+      {
         "command": "quarkusTools.buildBinary",
         "title": "Quarkus: Build executable"
       },
@@ -147,6 +151,11 @@
       {
         "command": "quarkusTools.deployToOpenShift",
         "title": "Quarkus: Deploy current Quarkus project to OpenShift (odo)"
+      },
+      {
+        "command": "quarkusTools.deployToOpenShift.short",
+        "title": "Deploy to OpenShift",
+        "icon": "$(rocket)"
       }
     ],
     "configuration": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -213,6 +214,13 @@
           "command": "quarkusTools.deployToOpenShift",
           "when": "quarkusProjectExistsOrLightWeight"
         }
+      ],
+      "editor/title": [
+        {
+          "command": "quarkusTools.deployToOpenShift",
+          "when": "quarkusProjectExistsOrLightWeight",
+          "group": "navigation@1"
```
rgrunber: Not sure if this particular `group` is needed.

rgrunber:
Should we also contribute the `Quarkus: Debug current Quarkus project` to the Run menu (beside the newly contributed Quarkus menu item) ?

~~Update : I also noticed we should probably restrict the deploy icon so it doesn't show up in any editor. Maybe only Java source files or application.properties files.~~

package.json
rgrunber: According to https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups , we should probably make this `1_run@1`

Update : `1_run` doesn't seem to display it as an icon (but in the `...` context menu), so sticking with `navigation`.

rgrunber:
quarkus-discoverable

rgrunber:
CC'ing @mohitsuman in case he may have some suggestions as well (regarding icons/command exposure,etc).

mohitsuman:
@rgrunber Can we use `OpenShift` icon in place of rocket icon. The icon can provide more intuitive way to identify it.

rgrunber:
Yes, this is a fairly straightforward change. The main issue is that the icon doesn't quite blend in with the other icons at the smaller size.

openshift-icon-titlebar

We could try to import https://rawgit.com/openshift/openshift-logos-icon/master/demo.html (icon-openshift) and see if that looks better integrated.

rgrunber:
openshift-icon-titlebar2

Thoughts on this vs. the one in previous comment ?

mohitsuman:
I think the last one looks much better and it also goes in alignment with the extension icon. So the icon means we lead them to the extension with the `Deploy to OpenShift` command.

rgrunber:
Merged as 2ca2a8179520ba701617d28293e2e2c0f836106e.

PR #414, opened 2021-11-12
Bump TypeScript version to ^4.1.0

AlexXuChen:
ts-essentials, a project dependency and dependency for monaco-page-objects, is on the 9.0.0 release and bumped to TypeScript@^4.1.0. Reference: https://github.com/krzkaczor/ts-essentials/blob/master/CHANGELOG.md

The local typescript version needs to be bumped to `^4.1.0` to allow recursive conditional types and avoid lint errors for dependencies.


AlexXuChen:
Fixed in #416

PR #417, opened 2021-11-18
Fix the npm-watch task

rgrunber:
- Remove the no longer available --info-verbosity flag (since updating
  webpack-cli)
- Remove problemWatcher since we can no longer track compilation without
  the verbosity flag
- Also use a JDK 11 for builds now that we've updated to Tycho 2.x


AlexXuChen (approved):
LGTM

PR #422, opened 2021-11-29
Fix support for HTTP for code.quarkus.io API endpoints

fbricon:
Fix http support to enable testing against a locally built code.quarkus.io endpoint. 
See https://github.com/quarkusio/code.quarkus.io/pull/680#issuecomment-981908963

﻿Signed-off-by: Fred Bricon <fbricon@gmail.com>

### DIFF
--- a/src/utils/codeQuarkusApiUtils.ts
+++ b/src/utils/codeQuarkusApiUtils.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { IncomingMessage } from "http";
+import * as http from "http";
 import * as https from "https";
 import * as yaml from "js-yaml";
 import * as path from "path";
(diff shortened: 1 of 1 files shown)


rgrunber:
Makes sense to me. Found a similar snippet in https://github.com/redhat-developer/vscode-xml/blob/master/src/server/binary/binaryServerStarter.ts#L265-L270 . Not sure why the test failed..

AlexXuChen:
I'm seeing an error thrown from `javaServerStarter.ts#49`:
Screenshot from 2021-11-29 14-15-29
Just checked and I didn't see this on master.

rgrunber:
> I'm seeing an error thrown from `javaServerStarter.ts#49`: Screenshot from 2021-11-29 14-15-29
Just checked and I didn't see this on master.

Did you pull the latest changes ? The error comes from the qute support and isn't related to this.

AlexXuChen:
> Did you pull the latest changes ? The error comes from the qute support and isn't related to this.

Ah okay, no longer seeing the error. However the local ui test is timing out, same as on the runner.

rgrunber:
Yup, I'm seeing the same locally without this change, so at least it isn't related. ~~My guess, some odd behaviour in newer VS Code release.~~

rgrunber:
I found the issue. It affects vscode-java when no workspace is defined, which is how we run our wizard tests. As soon as I push a (temporary) fix to the test settings, we can rebase this. When we release a new vscode-java, we can get rid of the fix.

https://github.com/redhat-developer/vscode-quarkus/pull/423

PR #424, opened 2021-12-10
Make clickable CodeLens data model

angelozerr:
Make clickable CodeLens data model

See https://github.com/redhat-developer/quarkus-ls/issues/471

requires https://github.com/redhat-developer/quarkus-ls/pull/480


angelozerr:
For the moment it doesn't work. I need to investigate what is the problem.

angelozerr:
It should work now to jump to method / field and method parameter. The only thing which doesn't work is to jump to data (method invocation of template.data).

fbricon:
Thanks @angelozerr!

PR #431, opened 2022-01-11
Add quarkus.tools.qute.validation.enabled in doc

angelozerr:
Add quarkus.tools.qute.validation.enabled in doc


angelozerr:
Fixed in https://github.com/redhat-developer/vscode-quarkus/pull/442

PR #434, opened 2022-01-12
Changed direct index access highlighting

### LINKED ISSUE #432: Direct index access highlighting looks bad
Accessing a list or array by index directly doesn't render properly:
Screenshot 2022-01-11 at 19 37 16

AlexXuChen:
Changed direct index access highlighting to be consistent with standard highlighting.
Screenshot from 2022-01-13 10-13-19

Fixes #432 

### DIFF
--- a/language-support/qute/qute.tmLanguage.json
+++ b/language-support/qute/qute.tmLanguage.json
@@ -481,7 +481,7 @@
               "name": "punctuation.separator.period.java"
             },
             "2": {
-              "name": "invalid.illegal.identifier.java"
+              "name": "constant.numeric.decimal.java"
             }
           }
         }


angelozerr:
IMHO I think it should be cleaner to

 * remove https://github.com/redhat-developer/vscode-quarkus/blob/ca94e697273828a80c9d5191c0ea9f89fa7e09a0/language-support/qute/qute.tmLanguage.json#L478
 * fix the match https://github.com/redhat-developer/vscode-quarkus/blob/ca94e697273828a80c9d5191c0ea9f89fa7e09a0/language-support/qute/qute.tmLanguage.json#L467 to support number.

angelozerr:
Or perhaps it should be better to highlight the number as number with `constant.numeric.decimal.java` ?

Because 

`items.get(0)` is the same things than `items.0`

image

@fbricon what do you think about that?

fbricon:
> Or perhaps it should be better to highlight the number as number with `constant.numeric.decimal.java` ?
> 
> Because
> 
> `items.get(0)` is the same things than `items.0`
> 
> image
> 
> @fbricon what do you think about that?

+1

angelozerr:
Thanks @AlexXuChen !

PR #437, opened 2022-01-14
Report telemetry data from the project generation wizard.

### LINKED ISSUE #405: Provide more granularity for the create project event
When collecting telemetry for the create project event, it'd be better to collect:
- the build system that was chosen
- whether starter code was selected

rgrunber:
- Fixes #405
- perform telemetry directly in the project generation wizard.
- Make deactivate() return an empty Promise to restore shutdown events

The project generation wizard has some cases where the `vscode.openFolder` command may be called. In those cases, the extension is shut down / disposed and the command never returns, so we aren't actually reporting this type of telemetry in certain cases! This means we can't do telemetry in the `registerCommandWithTelemetry` wrapper, and should do it just before the open folder logic.

Also, I'm not yet sure why, but having `deactivate()` return an empty Promise, restores the shutdown telemetry events for me, where before none were reported. Interestingly, the last reporting of a `shutdown` event for me happened on Dec 8th with VS Code 1.62.3. 1.63.0 was released the same day so it looks like the update might have changed something.

### DIFF
--- a/src/wizards/generateProject/generationWizard.ts
+++ b/src/wizards/generateProject/generationWizard.ts
@@ -8,13 +8,14 @@ import * as fse from 'fs-extra';
 import * as path from 'path';
 import { commands, OpenDialogOptions, QuickPickItem, Uri, window, workspace } from 'vscode';
 import { ZipFile } from 'yauzl';
-import { BuildToolName, INPUT_TITLE } from '../../definitions/constants';
+import { BuildToolName, INPUT_TITLE, VSCodeCommands } from '../../definitions/constants';
 import { ProjectGenState } from '../../definitions/inputState';
 import { QExtension } from '../../definitions/QExtension';
 import { QuarkusContext } from '../../QuarkusContext';
 import { CodeQuarkusFunctionality, PlatformVersionPickItem, getCodeQuarkusApiFunctionality, getDefaultFunctionality, getCodeQuarkusApiPlatforms } from '../../utils/codeQuarkusApiUtils';
 import { MultiStepInput, QuickPickParameters } from '../../utils/multiStepUtils';
 import { downloadProject } from '../../utils/requestUtils';
+import { CMD_SUCCEED_VALUE, sendTelemetry } from '../../utils/telemetryUtils';
 import { ExtensionsPicker } from './ExtensionsPicker';
 import { validateArtifactId, validateGroupId, validatePackageName, validateResourceName, validateVersion } from './validateInput';
 
@@ -23,7 +24,7 @@ import { validateArtifactId, validateGroupId, validatePackageName, validateResou
  *
  * This first part uses the helper class `MultiStepInput` that wraps the API for the multi-step case.
  */
-export async function generateProjectWizard() {
+export async function generateProjectWizard(): Promise<any> {
 
   let apiCapabilities: CodeQuarkusFunctionality;
   try {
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -75,7 +75,8 @@ export async function activate(context: ExtensionContext) {
   });
 }
 
-export function deactivate() {
+export async function deactivate(): Promise<void> {
+  return new Promise((_resolve, _reject) => {});
```
rgrunber: `Promise.resolve()`

src/wizards/generateProject/generationWizard.ts
```diff
@@ -202,6 +203,11 @@ export async function generateProjectWizard() {
   state.targetDir = await getTargetDirectory(state.artifactId);
 
   const projectGenState: ProjectGenState = state as ProjectGenState;
+  await sendTelemetry(VSCodeCommands.CREATE_PROJECT, {
```
fbricon: now 2 events are sent with this name, because an event is already sent on every command executed

rgrunber: This should be prevented by https://github.com/redhat-developer/vscode-quarkus/pull/437/files#diff-7aedfdde409c92096e09505c33a32ab6c61b36645e58da8a159ca63144f9c364R26 . When set to `true` (as done for `generateProjectWizard` registration), the success is not reported as it's assumed the command will handle it.

Although I just noticed we don't set `status` at this line, so this should definitely be fixed.

src/extension.ts
rgrunber: Ugh, so If I call `resolve()`, then this goes back to not reporting the shutdown. Looks like the shutdown only got reported because the promise would not get handled :\

I'm not sure why the shutdown isn't being reported. It happens the same in vscode-xml & vscode-quarkus.

Update : It could just be something specific to my machine so I'll just back out of this change for now.

src/wizards/generateProject/generationWizard.ts
```diff
@@ -202,6 +203,12 @@ export async function generateProjectWizard() {
...
+    status: CMD_SUCCEED_VALUE,
```
fbricon: status is set as succeeded even before we perform the actual project creation. So we'll never see any failure

rgrunber: Right. I think I can create a separate helper that does the `vscode.openFolder` and reports the telemetry right before that. That should ensure we still report failure if something happens in the download/extraction of the project.

rgrunber: @fbricon I've updated the PR to perform the telemetry reporting just before `openProject(..)` is called. There's just one issue :

If starting with a workspace that has no folders opened, and generating a Quarkus project from that, it still doesn't report the telemetry. I'm using `sendTelemetry(...).then(() => openProject(projectDir));` . I can see :

```
vscode-redhat-telemetry: Event received: quarkusTools.createProject
vscode-redhat-telemetry: Sending 'track' event with
{"userId":"...","event":"quarkusTools.createProject","properties":{"status":"succeeded","buildTool":"Maven","shouldGenerateCode":true,"extensions":"io.quarkus:quarkus-agroal,io.quarkiverse.amazonservices:quarkus-amazon-dynamodb,io.quarkiverse.amazonservices:quarkus-amazon-iam","extension_name":"redhat.vscode-quarkus","extension_version":"1.9.1","app_name":"Visual Studio Code","app_version":"1.63.2","app_kind":"Desktop","app_remote":false}...}
```
in the debug log, but the event never ends up recorded. Is there some part of the telemetry reporting that doesn't finish immediately and would get interrupted by the eventual `vscode.openFolder` call ?

rgrunber:
I think I'll merge this in for the upcoming release. There's a case where telemetry doesn't seem to be sent, but it's better to report some/most amounts than nothing at all.

PR #439, opened 2022-01-14
Support exclusion file for Qute for validation

angelozerr:
Support exclusion file for Qute for validation

requires https://github.com/redhat-developer/quarkus-ls/pull/540


angelozerr:
Close for https://github.com/redhat-developer/vscode-quarkus/pull/447

Issue #428, opened 2022-01-07
Improve syntax coloration for qute section


angelozerr:
It should be nice to have different color for qute core section like #if #for etc and custom user section like #body which is a section defined by the user with #include / #insertion

It should be nice to highlight the in inside a #for 

Given this qute template:

```
{#include base} 
  {#title}My Title{/title} 
  <div> 
    My body.
  </div>
{/include}
```

today we higlight like this:

![image](https://user-images.githubusercontent.com/1932211/150163353-3916e32d-8421-4ef3-9fb0-f6877a8c358c.png)

`#include` is a core Qute section and `#title` is a custom client section, so it should be very nice to higlight `title` with an another color:

![image](https://user-images.githubusercontent.com/1932211/150163488-cc94b8c5-ba0a-4928-b658-eb9ae6b85f1d.png)

angelozerr:
Fixed with https://github.com/redhat-developer/vscode-quarkus/pull/468

PR #444, opened 2022-01-19
Improved syntax coloration for Qute `in`

AlexXuChen:
Improved syntax coloration for Qute `in` keyword.
Screenshot from 2022-01-20 14-43-09

Part of #428 

### DIFF
--- a/language-support/qute/qute.tmLanguage.json
+++ b/language-support/qute/qute.tmLanguage.json
@@ -362,6 +362,14 @@
         {
           "match": "\\b(const|goto)\\b",
           "name": "keyword.reserved.java"
+        },
+        {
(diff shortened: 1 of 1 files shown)


angelozerr:
I fear that it's more complex because this in should be higlighted only in #for and after the second parameter.

Here a sample which show in action problem with in:

```
{in}

{#let in=123 }

{/let}

{#for item in in in in in in in in items}
	{item.name}
{/for}

```
image

angelozerr:
We need to clean some rule lie const, goto which are not in part of Qute, but we can do that in a separate PR.

angelozerr:
Thanks @AlexXuChen !

PR #447, opened 2022-01-20
Support exclusion file for Qute for validation

angelozerr:
Support exclusion file for Qute for validation

requires https://github.com/redhat-developer/quarkus-ls/pull/544


fbricon:
@angelozerr thanks, it works exactly as I expected. \o/

PR #449, opened 2022-01-21
Improved TextMate classification for Qute `include` tags

AlexXuChen:
Improved TextMate classification for Qute `include` tags.

Part of #428 

### DIFF
--- a/language-support/qute/qute.tmLanguage.json
+++ b/language-support/qute/qute.tmLanguage.json
@@ -77,7 +77,7 @@
 			"name": "string.unquoted.cdata.qute"
 		},
     "section_start_tag": {
-      "begin": "({)(#)(else\\sif|\\w+(\\.\\w+)*)",
+      "begin": "({)(#)(else\\sif|(\\w+(\\.\\w+)*)^(?!.*(return|break|case|continue|default|do|while|for|switch|if|else|include)))?",
       "end": "(\\/)?((?<![\\\\])})",
       "beginCaptures": {
         "1": {
@@ -362,6 +362,10 @@
         {
           "match": "\\b(const|goto)\\b",
           "name": "keyword.reserved.java"
+        },
+        {
+          "match": "\\b(include)\\b",
+          "name": "keyword.control.import"
         }
       ]
     },


language-support/qute/qute.tmLanguage.json
```diff
@@ -362,6 +362,10 @@
         {
           "match": "\\b(const|goto)\\b",
           "name": "keyword.reserved.java"
+        },
+        {
+          "match": "\\b(include)\\b",
+          "name": "keyword.control.import"
```
AlexXuChen, line 368: Currently there is no colour difference for every keyword/no notable different shade of pink to use.

language-support/qute/qute.tmLanguage.json
```diff
@@ -362,6 +362,10 @@
         {
           "match": "\\b(const|goto)\\b",
```
angelozerr, line 363: This pattern should be removed. const, goto, return etc are not a keyword, etc for Qute template.

language-support/qute/qute.tmLanguage.json
angelozerr: I think it's not teh correct fix, because if you write `{include}` it should be not a keyword.

language-support/qute/qute.tmLanguage.json
```diff
@@ -77,7 +77,7 @@
 			"name": "string.unquoted.cdata.qute"
 		},
     "section_start_tag": {
-      "begin": "({)(#)(else\\sif|\\w+(\\.\\w+)*)",
+      "begin": "({)(#)(else\\sif|(\\w+(\\.\\w+)*)^(?!.*(return|break|case|continue|default|do|while|for|switch|if|else|include)))?",
```
angelozerr, line 80: I think it's not the correct fix,because 

 * `{include}` should be not higligthed as keyword.
 * `{#include`  must be higlighted as keyword.

perhaps a correct fix is to have `section_start_tag` and `section_core_start_tag` (which higlight with,include,etc)

language-support/qute/qute.tmLanguage.json
AlexXuChen: If this fix is incorrect, then we shouldn't have https://github.com/redhat-developer/vscode-quarkus/blob/master/language-support/qute/qute.tmLanguage.json#L325 either, since syntax like for, if, etc.. is doesn't need to follow a `#` or `/` and is still a keyword.

AlexXuChen:
Fix in #468

PR #451, opened 2022-01-26
Force the proper Qute language for files under resources/templates

### LINKED ISSUE #450: Automatically force qute language of files under src/main/resources/templates of Quarkus projects
Nobody actually uses the qute.[html|txt|json|yaml] suffixes, especially in all the quarkus/qute examples. 
We should try to make our users' life easier by automatically switching to the proper language id, files under src/main/resources/templates of Quarkus projects

Similar to https://github.com/redhat-developer/vscode-quarkus/blob/master/src/extension.ts#L96

fbricon:
Fixes #450

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "languages": [
       {
         "id": "quarkus-properties",
+        
         "filenamePatterns": [
           "application.properties",
           "application-?*.properties"
--- a/package.json
+++ b/package.json
@@ -75,7 +75,6 @@
     "languages": [
       {
         "id": "quarkus-properties",
-        
         "filenamePatterns": [
           "application.properties",
           "application-?*.properties"
--- a/src/QuarkusConfig.ts
+++ b/src/QuarkusConfig.ts
@@ -31,6 +31,7 @@ export namespace QuarkusConfig {
   export const ALWAYS_SHOW_WELCOME_PAGE = 'quarkus.tools.alwaysShowWelcomePage';
   export const DEBUG_TERMINATE_ON_EXIT = 'quarkus.tools.debug.terminateProcessOnExit';
   export const PROPERTIES_LANGUAGE_MISMATCH = "quarkus.tools.propertiesLanguageMismatch";
+  export const QUARKUS_OVERRIDE_LANGUAGE_ID = 'quarkus.tools.override.languageId';
 
   export function getApiUrl(): string {
     if (process.env.VSCODE_QUARKUS_API_URL) {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -53,7 +53,7 @@ export async function activate(context: ExtensionContext) {
   // When a text document is opened,  update their language ID.
   context.subscriptions.push(
     workspace.onDidOpenTextDocument((document) => {
-      updateLanguageId(document, updatedDocumentsCache, false /* need the confirmation popup to STFU */);
+      updateLanguageId(document, updatedDocumentsCache, true);
     })
   );
   context.subscriptions.push(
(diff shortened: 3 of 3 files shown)


AlexXuChen (commented):
Looks like it's working as intended, given the following project directory:
```
src
├── main
    ├── other
        ├──a.html
    ├── resources
        ├── b.html
        ├── templates
            ├── nested_resource
                ├── c.html
            ├── d.html
```
Only `c.html` and `d.html` are recognized as `Qute HTML` files, while `a.html` and `b.html` are considered `HTML` files. This works `txt`, `yaml` and `json`, and doesn't apply to any unsupported file extension, such as `py`, `java`, etc..

src/extension.ts
```diff
@@ -112,30 +113,45 @@ function tryToForceLanguageId(document: TextDocument, fileName: string, document
...
+              + `to be '${languageId}' in order to provide language support. `
+              + `This behavior can be configured in settings.`, DISABLE_IN_SETTINGS, CONFIGURE_IN_SETTINGS);
+            response.then(result => {
+              if (result === CONFIGURE_IN_SETTINGS) {
+                commands.executeCommand('workbench.action.openSettings', QuarkusConfig.PROPERTIES_LANGUAGE_MISMATCH);
+              } else if (result === DISABLE_IN_SETTINGS) {
+                QuarkusConfig.setPropertiesLanguageMismatch(PropertiesLanguageMismatch.ignore).then(() => {
+                  languages.setTextDocumentLanguage(document, oldLanguageId);
+                });
+                context.globalState.update(QuarkusConfig.QUARKUS_SET_LANGUAGE_ID, 'true');
```
rgrunber: So if the user chooses `CONFIGURE_IN_SETTINGS`, the popup still gets shown every time ? Based on the name I figured it should stop showing in both cases.

src/extension.ts
```diff
@@ -112,30 +113,45 @@ function tryToForceLanguageId(document: TextDocument, fileName: string, document
         const oldLanguageId: string = document.languageId;
         languages.setTextDocumentLanguage(document, languageId);
         if (notifyUser && !documentCache.includes(document.fileName)) {
```
rgrunber, line 115: I'm confused about this line. I know it wasn't touched, but with this change `updateLanguageId(..)` always sets its last parameter as `false`. This means that `tryToForceLanguageId` will have the `notifyUser` parameter always `false`. So the question is, when does this code below run, if `notifyUser` is always `false`.

Am I missing some case ?

**Update** : I guess now that I think of it, I'm fine with this staying as long as we confirm we want to keep the functionality, even though it may not be used.

src/extension.ts
AlexXuChen: Should we make it more apparent with a "Don't Show Again" option?

rgrunber: No, I think it should be automatic. Whichever option you click, the popup shouldn't be shown again.

AlexXuChen:
The pop-up:
Screenshot from 2022-01-27 16-40-38

rgrunber:
Ok, just move `context.globalState.update(QuarkusConfig.QUARKUS_SET_LANGUAGE_ID, 'true');` [back to where it was](https://github.com/redhat-developer/vscode-quarkus/compare/f170558a19a1d615da3b20703425f5285a8366c4..93f09c457fc904b7a2e202f093e2e6b06173b08c). I realized that if we don't, then the `documentCache` logic becomes useless.

After that, I think this is ready to squash+merge.

src/extension.ts
```diff
@@ -162,5 +178,13 @@ async function updateLanguageId(document: TextDocument, documentCache: string[],
       return;
     }
     tryToForceLanguageId(document, fileName, documentCache, propertiesLanguageMismatch, 'yaml', onExtensionLoad);
+  } else if (document.fileName.includes(`resources${path.sep}templates${path.sep}`)) {
```
AlexXuChen, line 181: After testing, it appears this is platform independent as required

package.json
```diff
@@ -75,6 +75,7 @@
     "languages": [
       {
         "id": "quarkus-properties",
+
```
fbricon: unnecessary change

src/QuarkusConfig.ts
```diff
@@ -31,6 +31,7 @@ export namespace QuarkusConfig {
   export const ALWAYS_SHOW_WELCOME_PAGE = 'quarkus.tools.alwaysShowWelcomePage';
   export const DEBUG_TERMINATE_ON_EXIT = 'quarkus.tools.debug.terminateProcessOnExit';
   export const PROPERTIES_LANGUAGE_MISMATCH = "quarkus.tools.propertiesLanguageMismatch";
+  export const QUARKUS_SET_LANGUAGE_ID = 'quarkus.tools.set.language.id';
```
fbricon: QUARKUS_OVERRIDE_LANGUAGE_ID = 'quarkus.tools.override.languageId';

PR #453, opened 2022-01-31
Disabled Qute language id pop-up after first appearance

AlexXuChen:
Disabled Qute language id pop-up after first appearance and removed `documentCache`

Fix for #451 

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -45,22 +45,14 @@ export async function activate(context: ExtensionContext) {
     })
   );
 
-  let updatedDocumentsCache: string[] = [];
   // When extension is started, loop for each text documents which are opened to update their language ID.
   workspace.textDocuments.forEach(document => {
-    updateLanguageId(document, updatedDocumentsCache, false);
+    updateLanguageId(document, false);
   });
   // When a text document is opened,  update their language ID.
   context.subscriptions.push(
     workspace.onDidOpenTextDocument((document) => {
-      updateLanguageId(document, updatedDocumentsCache, true);
-    })
-  );
-  context.subscriptions.push(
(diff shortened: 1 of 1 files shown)


rgrunber:
LGTM. I wouldn't mind confirming but this should disable the popup after the first time.

AlexXuChen:
> LGTM. I wouldn't mind confirming but this should disable the popup after the first time.

Yep, working as intended:
Peek 2022-01-31 11-51

Something I noticed however: I was debugging workspaceTrust, and after disabling and re-enabling to test a vscode-xml fix, the `quarkus.tools.propertiesLanguageMismatch` was set to `ignore`, even though the default is `forceQuarkus`, so the language id was never forced. I never configured the setting, and even cleared my workspace cache before testing the untrusted workspace, so it seems unusual for this to happen. Thoughts?

rgrunber:
> > LGTM. I wouldn't mind confirming but this should disable the popup after the first time.
> 
> Something I noticed however: I was debugging workspaceTrust, and after disabling and re-enabling to test a vscode-xml fix, the `quarkus.tools.propertiesLanguageMismatch` was set to `ignore`, even though the default is `forceQuarkus`, so the language id was never forced. I never configured the setting, and even cleared my workspace cache before testing the untrusted workspace, so it seems unusual for this to happen. Thoughts?

Then let's merge this.

`Disable Language Updating` will set it to ignore https://github.com/redhat-developer/vscode-quarkus/blob/master/src/extension.ts#L128 .

AlexXuChen:
> `Disable Language Updating` will set it to ignore https://github.com/redhat-developer/vscode-quarkus/blob/master/src/extension.ts#L128 .

Ah so I must have disabled it once.

PR #454, opened 2022-02-01
Document force qute on type mismatch behavior

fbricon:

### DIFF
--- a/docs/qute/TemplateSupport.md
+++ b/docs/qute/TemplateSupport.md
@@ -1,8 +1,8 @@
 # Qute Template support
 
-Qute support provides `completions`, `validation`, `hover`, `definition`, `highlight`, `linked editing` in Qute template.
+Qute support provides `completions`, `validation`, `hover`, `definition`, `highlight`, `linked editing` in Qute templates.
 
-To benefit with this support, we recommend your template files to follow the `*.qute.*` naming pattern (ex: `items.qute.html`, `items.qute.txt`, `items.qute.yaml`, etc.).
+To benefit from this support, HTML, Text, YAML and JSON files in `**/resources/templates/**` are automatically opened as Qute templates. You can disable that behavior by setting the `quarkus.tools.propertiesLanguageMismatch` property to `ignore`, or `prompt` if you prefer being asked to change the type when opening the file. 
 
 ## Parameter declaration support
 


rgrunber (approved):
LGTM.

docs/qute/TemplateSupport.md
```diff
@@ -1,105 +1,105 @@
...
-
-In this case, you don't need to declare parameter declaration to benefit from Qute completion, validation, etc. of the Java data model.
-
-CodeLens are clickable to jump to the proper Java type, method or field.
\ No newline at end of file
+# Qute Template support
+
+Qute support provides `completions`, `validation`, `hover`, `definition`, `highlight`, `linked editing` in Qute templates.
+
+To benefit from this support, HTML, Text, YAML and JSON files in `**/resources/templates/**` are automatically opened as Qute templates. You can disable that behavior by setting the `quarkus.tools.propertiesLanguageMismatch` property to `ignore`, or `prompt`, if you prefer being asked to change the type when opening the file.
```
AlexXuChen: ...to `ignore`, or `prompt` if you prefer... (no comma)

AlexXuChen (changes requested):

docs/qute/TemplateSupport.md
rgrunber: Is this a rule ? I think a reader naturally pauses after "prompt", and before "if you prefer ..."

AlexXuChen: The "or" begins an independent clause (since it's separate from the `ignore` when separated by a comma), which is also proceeded by a description. If I were to describe something, I wouldn't put a comma, (i.e. "`prompt` is the option to ask the user if they prefer being asked to change the type when opening the file").

rgrunber: Ah right. the description describes  `prompt` and not the option in general. Ok, then sure, the comma shouldn't be there :P

AlexXuChen: Sorry if this is nit-picky, just a small adjustment

fbricon: geeez ;-)

fbricon: Fixed

AlexXuChen (approved):
LGTM

PR #455, opened 2022-02-01
Added toggle button for enable/disable Qute validation

### LINKED ISSUE #448: Toggle button on Qute editor to enable/disable validation
It should be nice to provide a toggle button on the the top right of the editor to enable/disable validation. 

We need to know the validation state (eabled or disabled?) of the opened template, and it will require to implement a custom command on Qute LS side. The state of the button will be synchronized with the vaidation state.

Click on button will active/desactive the validation for the Qute template. In other words it will update:

 * qute.validation.enabled to true in the workspace folder scope
 * remove the template from qute.validation.excluded pattern.

AlexXuChen:
Added toggle button for enable/disable Qute validation.
qutevalidationtemplatebutton

Fixes #448 

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,6 +159,16 @@
         "command": "quarkusTools.deployToOpenShift.short",
         "title": "Deploy to OpenShift",
         "icon": "icons/openshift.png"
+      },
+      {
+        "command": "qute.validation.enabled.toggle.off",
+        "title": "Disable Qute Validation",
+        "icon": "$(eye-watch)"
+      },
+      {
+        "command": "qute.validation.enabled.toggle.on",
+        "title": "Enable Qute Validation",
+        "icon": "$(eye-closed)"
       }
     ],
     "configuration": {
@@ -250,6 +260,14 @@
         {
           "command": "quarkusTools.deployToOpenShift",
           "when": "quarkusProjectExistsOrLightWeight"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.off",
+          "when": "false"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.on",
+          "when": "false"
         }
       ],
       "editor/title": [
@@ -257,6 +275,16 @@
           "command": "quarkusTools.deployToOpenShift.short",
           "when": "quarkusProjectExistsOrLightWeight && resourceExtname == .java",
           "group": "navigation@10"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.off",
+          "when": "quarkusProjectExistsOrLightWeight && editorLangIdSupportsQute && editorQuteValidationEnabled",
+          "group": "navigation@9"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.on",
+          "when": "quarkusProjectExistsOrLightWeight && editorLangIdSupportsQute && !editorQuteValidationEnabled",
+          "group": "navigation@9"
         }
       ],
       "editor/title/run": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,10 +49,12 @@ export async function activate(context: ExtensionContext) {
   workspace.textDocuments.forEach(document => {
     updateLanguageId(document, false);
   });
-  // When a text document is opened,  update their language ID.
+  // When a text document is opened, update their language ID.
   context.subscriptions.push(
-    workspace.onDidOpenTextDocument((document) => {
-      updateLanguageId(document, true);
+    workspace.onDidOpenTextDocument(async (document) => {
+      if (!(document.uri.scheme === 'git')) {
+        await updateLanguageId(document, true);
+      }
     })
   );
 
--- a/src/qute/commands/commandConstants.ts
+++ b/src/qute/commands/commandConstants.ts
@@ -25,6 +25,20 @@ export namespace QuteClientCommandConstants {
    */
   export const COMMAND_CONFIGURATION_UPDATE = 'qute.command.configuration.update';
 
+  /**
+   * Toggle on extension for Qute validation setting.
+   */
+  export const QUTE_VALIDATION_ENABLED_TOGGLE_ON = 'qute.validation.enabled.toggle.on';
+
+  /**
+   * Toggle off extension for Qute validation setting.
+   */
+  export const QUTE_VALIDATION_ENABLED_TOGGLE_OFF = 'qute.validation.enabled.toggle.off';
+
+  /**
+  * Client command to execute an XML command on XML Language Server side.
+  */
+  export const EXECUTE_WORKSPACE_COMMAND = 'qute.workspace.executeCommand';
 }
 
 /**
@@ -36,6 +50,11 @@ export namespace QuteServerCommandConstants {
    * Qute LS Server command to generate template content on the Qute Language Server
    */
   export const GENERATE_TEMPLATE_CONTENT = 'qute.command.generate.template.content';
+
+  /**
+   * Check if a Qute template is excluded from validation.
+   */
+  export const QUTE_VALIDATION_TEMPLATE_STATUS = 'qute.command.validation.template.status';
 
 }
 
--- a/src/qute/commands/registerCommands.ts
+++ b/src/qute/commands/registerCommands.ts
@@ -1,8 +1,10 @@
 import { TextEncoder } from "util";
-import { commands, ConfigurationTarget, ExtensionContext, Position, Range, Selection, Uri, window, workspace, WorkspaceConfiguration } from "vscode";
+import { commands, ConfigurationTarget, ExtensionContext, Position, Range, Selection, TextDocument, TextEditor, Uri, window, workspace, WorkspaceConfiguration } from "vscode";
 import { ConfigurationItem, Location } from "vscode-languageclient";
 import { QuteSettings } from "../languageServer/settings";
 import { QuteClientCommandConstants, QuteJdtLsServerCommandConstants, QuteServerCommandConstants } from "./commandConstants";
+import { CancellationToken, ExecuteCommandParams, ExecuteCommandRequest } from "vscode-languageclient";
+import { LanguageClient } from "vscode-languageclient/node";
 
 /**
  * Register custom vscode command for Qute support.
@@ -14,6 +16,29 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
   registerGenerateTemplateFileCommand(context);
   registerJavaDefinitionCommand(context);
   registerConfigurationUpdateCommand(context);
+  registerQuteValidationToggleCommand(context);
+}
+
+export function registerQuteExecuteWorkspaceCommand(context: ExtensionContext, languageClient: LanguageClient) {
+  // Register client command to execute custom Qute Language Server command
+  context.subscriptions.push(commands.registerCommand(QuteClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, (command, ...rest) => {
+    let token: CancellationToken;
+    let commandArgs: any[] = rest;
+    if (rest && rest.length && CancellationToken.is(rest[rest.length - 1])) {
+      token = rest[rest.length - 1];
+      commandArgs = rest.slice(0, rest.length - 1);
+    }
+    const params: ExecuteCommandParams = {
+      command,
+      arguments: commandArgs
+    };
+    if (token) {
+      return languageClient.sendRequest(ExecuteCommandRequest.type, params, token);
+    }
+    else {
+      return languageClient.sendRequest(ExecuteCommandRequest.type, params);
+    }
+  }));
 }
 
 /**
@@ -70,6 +95,55 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
 }
 
 /**
+ * Toggle for enabling/disabling Qute validation.
+ *
+ * @param context
+ */
+function registerQuteValidationToggleCommand(context: ExtensionContext) {
+  // Toggle off - add file to exclusion
+  context.subscriptions.push(commands.registerCommand(QuteClientCommandConstants.QUTE_VALIDATION_ENABLED_TOGGLE_OFF, async (uri?: Uri) => {
+    // Add the editor document file uri to the `qute.validation.exclusion` array
+    const templateUri = uri.toString();
+    const edit = {
+      scopeUri: templateUri,
+      value: templateUri,
+      editType: ConfigurationItemEditType.Add,
+      section: QuteSettings.QUTE_VALIDATION_EXCLUDED
+    } as ConfigurationItemEdit;
+    await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    await commands.executeCommand('setContext', 'editorQuteValidationEnabled', false);
+  }));
+
+  // Toggle on - remove file(s) from exclusion
+  context.subscriptions.push(commands.registerCommand(QuteClientCommandConstants.QUTE_VALIDATION_ENABLED_TOGGLE_ON, async (uri?: Uri) => {
+    const templateUri = uri.toString();
+    // Consume `qute.command.validation.template.status` command from qute-ls to retrieve files excluded from validation
+    const result: TemplateValidationStatus = await commands.executeCommand(QuteServerCommandConstants.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
+    // Enable the workspace `qute.validation.enabled` setting
+    if (!result.validationEnabled) {
+      const edit = {
+        scopeUri: templateUri,
+        value: true,
+        editType: ConfigurationItemEditType.Update,
+        section: QuteSettings.QUTE_VALIDATION_ENABLED
+      } as ConfigurationItemEdit;
+      await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    }
+    // Remove the editor document file uri to the `qute.validation.exclusion` array if it exists
+    if (result.excluded.length > 0) {
+      const edit = {
+        scopeUri: templateUri,
+        value: result.excluded,
+        editType: ConfigurationItemEditType.Remove,
+        section: QuteSettings.QUTE_VALIDATION_EXCLUDED
+      } as ConfigurationItemEdit;
+      await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    }
+    await commands.executeCommand('setContext', 'editorQuteValidationEnabled', true);
+  }));
+}
+
+/**
  * Update a given setting from the Qute language server.
  *
  * @param context the extension context.
@@ -77,7 +151,7 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
 export function registerConfigurationUpdateCommand(context: ExtensionContext) {
   context.subscriptions.push(commands.registerCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, async (configItemEdit: ConfigurationItemEdit) => {
     const section = configItemEdit.section;
-    const value = getSettingsValue(configItemEdit.value, configItemEdit.section);
+    const value = getSettingsValue(configItemEdit.value, configItemEdit.section, configItemEdit.editType);
     const config = getConfiguration(configItemEdit.scopeUri);
     switch (configItemEdit.editType) {
       case ConfigurationItemEditType.Add:
@@ -91,17 +165,53 @@ export function registerConfigurationUpdateCommand(context: ExtensionContext) {
         config.workspaceConfiguration.update(section, value, config.target);
         break;
       }
+      case ConfigurationItemEditType.Remove: {
+        removeFromPreferenceArray(config, section, value);
+        break;
+      }
     }
   }));
 }
 
-function getSettingsValue(value: any, section: string) {
+/**
+   * Sets the `editorQuteValidationEnabled` based on `qute.command.validation.template.status` and `qute.validation.enabled`.
+   */
+async function checkQuteValidationFromExclusionContext(uri: Uri) {
+  const templateUri = uri.toString();
+  const result: TemplateValidationStatus = await commands.executeCommand(QuteClientCommandConstants.EXECUTE_WORKSPACE_COMMAND, QuteServerCommandConstants.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
+  await commands.executeCommand('setContext', 'editorQuteValidationEnabled', (result.validationEnabled && (!result.excluded || result.excluded.length === 0)));
+}
+
+/**
+   * Synchronize Qute validation button from the activated editor:
+   *
+   * - the button can be hidden (when file is not a Qute template file).
+   * - the button can be shown with eye-closed (when file is a Qute template file and validation is enabled).
+   * - the button can be shown with eye-opened (when file is a Qute template file and validation is disabled)
+   */
+ export async function synchronizeQuteValidationButton(editor: TextEditor) {
+  const document = editor?.document;
+  if (!document) {
+    return;
+  }
+  // Is the document is a Qute template file?
+  const quteSupported = document.languageId.includes('qute');
+  await commands.executeCommand('setContext', 'editorLangIdSupportsQute', quteSupported);
+  if (quteSupported) {
+    // Here the button will be visible, we should display it with eye-closed icon (validation is enabled) or with eye-opened icon (validation is disabled).
+    await checkQuteValidationFromExclusionContext(document.uri);
+  }
+}
+
+function getSettingsValue(value: any, section: string, editType: ConfigurationItemEditType) {
   switch (section) {
     case QuteSettings.QUTE_VALIDATION_EXCLUDED:
-      return workspace.asRelativePath(Uri.parse(value), false);
-    default:
-      return value;
+      if (editType === ConfigurationItemEditType.Add) {
+        return workspace.asRelativePath(Uri.parse(value), false);
+      }
+      break;
   }
+  return value;
 }
 
 interface IConfiguration {
@@ -128,10 +238,36 @@ function getConfiguration(scopeUri: string): IConfiguration {
 function addToPreferenceArray<T>(config: IConfiguration, key: string, value: T): void {
   const workspaceConfiguration = config.workspaceConfiguration;
   const configArray: T[] = workspaceConfiguration.get<T[]>(key, []);
-  if (configArray.includes(value)) {
-    return;
+  if (Array.isArray(value)) {
+    value.forEach(item => {
+      if (!configArray.includes(item)) {
+        configArray.push(item);
+      }
+    });
+  } else {
+    if (configArray.includes(value)) {
+      return;
+    }
+    configArray.push(value);
   }
-  configArray.push(value);
+  workspaceConfiguration.update(key, configArray, config.target);
+}
+
+function removeFromPreferenceArray<T>(config: IConfiguration, key: string, value: T): void {
+  const workspaceConfiguration = config.workspaceConfiguration;
+  const configArray: T[] = workspaceConfiguration.get<T[]>(key, []);
+  if (Array.isArray(value)) {
+    value.forEach(item => {
+      if (configArray.includes(item)) {
+        configArray.splice(configArray.indexOf(item), 1);
+      }
+    });
+  } else {
+    if (!configArray.includes(value)) {
+      return;
+    }
+    configArray.splice(configArray.indexOf(value), 1);
+  }
   workspaceConfiguration.update(key, configArray, config.target);
 }
 
@@ -143,5 +279,11 @@ interface ConfigurationItemEdit extends ConfigurationItem {
 enum ConfigurationItemEditType {
   Add = 0,
   Delete = 1,
-  Update = 2
+  Update = 2,
+  Remove = 3
+}
+
+interface TemplateValidationStatus {
+  validationEnabled: boolean;
+  excluded: string[];
 }
--- a/src/qute/languageServer/client.ts
+++ b/src/qute/languageServer/client.ts
@@ -4,7 +4,7 @@ import { DidChangeConfigurationNotification, LanguageClientOptions } from 'vscod
 import { LanguageClient } from 'vscode-languageclient/node';
 import { ExtensionContext, commands, workspace, window, ConfigurationTarget, WorkspaceConfiguration } from 'vscode';
 import { prepareExecutable } from './javaServerStarter';
-import { registerVSCodeQuteCommands } from '../commands/registerCommands';
+import { registerQuteExecuteWorkspaceCommand, registerVSCodeQuteCommands, synchronizeQuteValidationButton } from '../commands/registerCommands';
 import { QuteClientCommandConstants } from '../commands/commandConstants';
 import { QuteSettings } from './settings';
 
@@ -44,8 +44,12 @@ export function connectToQuteLS(context: ExtensionContext) {
       },
       middleware: {
         workspace: {
-          didChangeConfiguration: () => {
+          didChangeConfiguration: async () => {
+            // A settings.json is updated:
+            // 1. send the new Quet settings to the Qute language server
             quteLanguageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: getQuteSettings() });
+            // 2. synchronize the Qute toggle button for validation
+            await synchronizeQuteValidationButton(window.activeTextEditor);
           }
         }
       }
@@ -66,6 +70,7 @@ export function connectToQuteLS(context: ExtensionContext) {
     const serverOptions = prepareExecutable(requirements);
     const quteLanguageClient = new LanguageClient('qute', 'Qute Support', serverOptions, clientOptions);
     context.subscriptions.push(quteLanguageClient.start());
+
     return quteLanguageClient.onReady().then(async () => {
       bindQuteRequest('qute/template/project');
       bindQuteRequest('qute/template/projectDataModel');
@@ -76,11 +81,31 @@ export function connectToQuteLS(context: ExtensionContext) {
       bindQuteRequest('qute/java/diagnostics');
       bindQuteRequest('qute/java/documentLink');
       bindQuteNotification('qute/dataModelChanged');
+
+      registerQuteExecuteWorkspaceCommand(context, quteLanguageClient);
+      // Refresh the Qute context when editor tab has the focus
+      context.subscriptions.push(
+        window.onDidChangeActiveTextEditor(async editor => {
+          await synchronizeQuteValidationButton(editor);
+        })
+      );
+      // Refresh the Qute context when the language id changed (HTML -> Qute HTML or Qute HTML -> HTML)
+      context.subscriptions.push(
+        workspace.onDidOpenTextDocument(async (document) => {
+          // when settings.json is updated, onDidOpenTextDocument is called,
+          // the Qute context must be refreshed only for the activate text editor.
+          if (window.activeTextEditor?.document === document) {
+            await synchronizeQuteValidationButton(window.activeTextEditor);
+          }
+        })
+      );
+      await setQuteValidationEnabledContext();
+      await synchronizeQuteValidationButton(window.activeTextEditor);
+
       if (!hasShownQuteValidationPopUp(context)) {
         await showQuteValidationPopUp(context);
       }
-    }
-    );
+    });
   });
 }
 
@@ -155,4 +180,12 @@ async function showQuteValidationPopUp(context: ExtensionContext) {
     workspace.getConfiguration().update(QuteSettings.QUTE_VALIDATION_ENABLED, true, ConfigurationTarget.Global);
   }
   context.globalState.update(QuteSettings.EXPERIMENTAL_QUTE_VALIDATION_FLAG, 'true');
+}
+
+/**
+   * Sets the `editorQuteValidationEnabled` context to `true` if the `qute.validation.enabled`
+   * setting is set to true. Sets to `false` otherwise.
+   */
+export async function setQuteValidationEnabledContext() {
+  await commands.executeCommand('setContext', 'editorQuteValidationEnabled', workspace.getConfiguration().get(QuteSettings.QUTE_VALIDATION_ENABLED));
 }
--- a/src/qute/languageServer/settings.ts
+++ b/src/qute/languageServer/settings.ts
@@ -14,7 +14,7 @@ export namespace QuteSettings {
   export const QUTE_VALIDATION_ENABLED = 'qute.validation.enabled';
 
   /**
- * Disable Qute validation for the given file name patterns settings.
- */
+   * Disable Qute validation for the given file name patterns settings.
+   */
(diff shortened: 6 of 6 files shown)


src/qute/commands/registerCommands.ts
```diff
@@ -13,6 +14,19 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
   registerGenerateTemplateFileCommand(context);
   registerJavaDefinitionCommand(context);
   registerConfigurationUpdateCommand(context);
+
+  context.subscriptions.push(commands.registerCommand(QuteSettings.QUTE_VALIDATION_ENABLED + QuteSettings.TOGGLE_ON, ()=> {
+    // workspace.getConfiguration().update(QuteSettings.QUTE_VALIDATION_ENABLED, false);
+    workspace.getConfiguration().update(QuteSettings.QUTE_VALIDATION_ENABLED, !workspace.getConfiguration().get(QuteSettings.QUTE_VALIDATION_ENABLED));
```
AlexXuChen: I'm aware of https://github.com/redhat-developer/vscode-quarkus/issues/448#issuecomment-1018703205, for now it's a PoC that the option can be toggled.

src/qute/languageServer/client.ts
```diff
@@ -79,6 +79,7 @@ export function connectToQuteLS(context: ExtensionContext) {
       if (!hasShownQuteValidationPopUp(context)) {
         await showQuteValidationPopUp(context);
       }
+      await setQuteValidationEnabledContext();
```
AlexXuChen: Need to add this to a listener so there is a check each time the button state changes (based on `quteValidationEnabled`)

package.json
```diff
@@ -248,6 +258,16 @@
           "command": "quarkusTools.deployToOpenShift.short",
           "when": "quarkusProjectExistsOrLightWeight && resourceExtname == .java",
           "group": "navigation@10"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.on",
+          "when": "quarkusProjectExistsOrLightWeight && quteValidationEnabled",
```
AlexXuChen: Need a way to identify if a Qute file is open

package.json
```diff
@@ -159,6 +159,16 @@
...
+      },
+      {
+        "command": "qute.validation.enabled.toggle.on",
+        "title": "Qute Validation Enabled",
+        "icon": "$(star-full)"
+      },
+      {
+        "command": "qute.validation.enabled.toggle.off",
+        "title": "Qute Validation Disabled",
+        "icon": "$(star-empty)"
```
rgrunber: Is there another set of 2 icons that better represent validation ? I was thinking of maybe search/search-stop but that could too easily be confused as some kind of search :\

**Update** : Maybe pass & pass-filled ?

https://code.visualstudio.com/api/references/icons-in-labels

AlexXuChen: Alternatively circle-large-outline and pass-filled?

package.json
```diff
@@ -248,6 +258,16 @@
...
           "group": "navigation@10"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.on",
+          "when": "quarkusProjectExistsOrLightWeight && editorLangId == qute-html && quteValidationEnabled",
+          "group": "navigation@9"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.off",
+          "when": "quarkusProjectExistsOrLightWeight && editorLangId == qute-html && !quteValidationEnabled",
```
fbricon: need to check editorLangId against all 4 of the qute flavors (html, txt, yaml, json)

package.json
```diff
@@ -159,6 +159,16 @@
...
+      },
+      {
+        "command": "qute.validation.enabled.toggle.on",
+        "title": "Qute Validation Enabled",
+        "icon": "$(pass-filled)"
+      },
+      {
+        "command": "qute.validation.enabled.toggle.off",
+        "title": "Qute Validation Disabled",
+        "icon": "$(circle-large-outline)"
```
fbricon: I like ${eye-watch} and ${eye-closed}

qute-validation-toggle

src/qute/languageServer/client.ts
```diff
@@ -156,3 +157,11 @@ async function showQuteValidationPopUp(context: ExtensionContext) {
   }
   context.globalState.update(QuteSettings.EXPERIMENTAL_QUTE_VALIDATION_FLAG, 'true');
 }
+
+/**
+   * Sets the `quteValidationEnabled` context to `true` if current workspace
+   * contains a Quarkus project. Sets to `false` otherwise.
```
fbricon: Description seems bogus

angelozerr:
@AlexXuChen please see my POC button https://github.com/redhat-developer/vscode-quarkus/pull/456 which consumes there https://github.com/redhat-developer/vscode-quarkus/pull/456/files#diff-1a68e0e2bba23b8f7afb6fa4e970ec5b00064bfb0a24c0f0c8c4c2b85740ae90R24 he Qute LS command to  know the validation status for a template 

This command returns data like https://github.com/redhat-developer/quarkus-ls/blob/master/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/commands/validation/TemplateValidationStatus.java

Thos command returns the reason why the validation for a template is disabled or not. It will give you the capability to update exclude patterns when a Qute template valiation must be enabled and it is defined in excluded.

My code is ugly, so please don't use it, but it give you some help (I hope) to consume the Qute LS command to know the validation status.

AlexXuChen:
> @AlexXuChen please see my POC button #456 which consumes there https://github.com/redhat-developer/vscode-quarkus/pull/456/files#diff-1a68e0e2bba23b8f7afb6fa4e970ec5b00064bfb0a24c0f0c8c4c2b85740ae90R24 he Qute LS command to know the validation status for a template
> 
> This command returns data like https://github.com/redhat-developer/quarkus-ls/blob/master/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/commands/validation/TemplateValidationStatus.java
> 
> Thos command returns the reason why the validation for a template is disabled or not. It will give you the capability to update exclude patterns when a Qute template valiation must be enabled and it is defined in excluded.
> 
> My code is ugly, so please don't use it, but it give you some help (I hope) to consume the Qute LS command to know the validation status.

@angelozerr I believe I already consume the command you mentioned: https://github.com/redhat-developer/vscode-quarkus/pull/455/files#diff-1a68e0e2bba23b8f7afb6fa4e970ec5b00064bfb0a24c0f0c8c4c2b85740ae90R81 I just need to combine with the `qute.command.validation.template.status` command.

package.json
```diff
@@ -257,6 +267,16 @@
           "command": "quarkusTools.deployToOpenShift.short",
           "when": "quarkusProjectExistsOrLightWeight && resourceExtname == .java",
           "group": "navigation@10"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.on",
```
rgrunber: After you take into account the file's validation state, I think it should be possible to merge these 2 commands into a single toggle command, along with registering a single command.

If the user toggles the validation, you'll first access the current validation state of the file (using the command Angelo has provided). Once you know that, you'll know which way to change the state, and that can all be done in one command.

AlexXuChen: Is there a way to assign 2 icons to a single command? Is there also a way from hiding the command from the command palette?

rgrunber: Nope, nevermind, can't be done. A menu is tied to a command, not to an icon.

src/qute/commands/registerCommands.ts
```diff
@@ -95,6 +116,15 @@ export function registerConfigurationUpdateCommand(context: ExtensionContext) {
   }));
 }
 
+/**
+   * Sets the `quteLanguageSupported` context to `true` if the editor language id
+   * is a qute template. Sets to `false` otherwise.
+   */
+ export async function checkQuteLanguageSupportedContext() {
+  const isSupportedOpen = window.activeTextEditor.document.languageId.includes('qute')
```
rgrunber: Would it be possible to pass in the `document` as a parameter here from the `onDidOpenTextDocument` call. It would avoid having to fetch it through the UI element. It shouldn't fail this way, but passing it in seems safer.

AlexXuChen: It seems when using the `document` param from `onDidOpenTextDocument`, the button will appear/disappear each time a file is opened. It seems each time a text document is opened, the first run of the listener returns true for `checkQuteLanguageContext`, then it will run again and return false and the button will disappear.

rgrunber: And you added `await` to the `updateLanguageId` call right ? Still doesn't work ?

rgrunber: And you added `await` to the `updateLanguageId` call right ? Still doesn't work ?

**Update** : Also just noticed you're missing a semicolon here.

AlexXuChen: Yes, same result:
Peek 2022-02-03 13-42

rgrunber: Ok I think I know why this is happening.

I just added a breakpoint on `isSupportedOpen` of `checkQuteLanguageSupportedContext`. The callback `onDidOpenTextDocument` is triggered twice. Once for the file, and a second time on the same file with `.git` as suffix. I've seen this before, but no idea why the extra event.

Let's leave it as-is and maybe I can investigate why that happens.

rgrunber: See https://github.com/microsoft/vscode/issues/131282 . Probably some git capabilities are generating the extra file. I guess we could technically have a check to ignore documents where the uri scheme is git.

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +82,40 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
+    const templateUri = uri.toString();
+    // from QuteTemplateValidationStatusCommandHandler
+    const result = await commands.executeCommand(QuteSettings.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
+    const edit = {
+      value: templateUri,
+      editType: ConfigurationItemEditType.Delete,
+      section: QuteSettings.QUTE_VALIDATION_EXCLUDED
+    } as ConfigurationItemEdit;
+    await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    await commands.executeCommand('setContext', 'quteValidationEnabled', !(result as any).validationEnabled);
```
angelozerr, line 115: You must check too that result.excluded is empty

angelozerr (changes requested):

src/qute/commands/registerCommands.ts
```diff
@@ -42,6 +43,18 @@ function registerGenerateTemplateFileCommand(context: ExtensionContext) {
   }));
 }
 
+// /**
```
angelozerr: Remove this commented code

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +82,40 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
+ *
+ * @param context
+ */
+function registerQuteValidationToggleCommand(context: ExtensionContext) {
+  // Toggle on
+  context.subscriptions.push(commands.registerCommand(QuteSettings.QUTE_VALIDATION_ENABLED + QuteSettings.TOGGLE_ON, async (uri?: Uri) => {
+    const templateUri = uri.toString();
+    // from QuteTemplateValidationStatusCommandHandler
+    const result = await commands.executeCommand(QuteSettings.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
+    const edit = {
```
angelozerr, line 107: It's not like this how he should work.

At first I suggest that you define an interface like this:

```ts
interface TemplateValidationStatus {
	validationEnabled: boolean;
	excluded: [string];
}
```

And the code should look like this:

```ts
    const result: TemplateValidationStatus = await commands.executeCommand(QuteServerCommandConstants.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
    if (!result.validationEnabled) {
      // The current settings 'is 'qute.validation.enabled' is set to false, update it to true
      const edit = {
        scopeUri: templateUri,
        value: true,
        editType: ConfigurationItemEditType.Update,
        section: QuteSettings.QUTE_VALIDATION_ENABLED
      } as ConfigurationItemEdit;
      await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
    }
    if (result.excluded.length > 0) {
      // The current settings 'is 'qute.validation.excluded' have some patterns which matches the template uri
      // Those patterns must be removed
      const edit = {
        scopeUri: templateUri,
        value: result.excluded,
        editType: ConfigurationItemEditType.UpdateByRemoving, // TODO : implement UpdateByRemoving which must remove the given value (excluded array) from the existing qute.validation.excluded
        section: QuteSettings.QUTE_VALIDATION_EXCLUDED
      } as ConfigurationItemEdit;
      await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
    }
    await commands.executeCommand('setContext', 'quteValidationEnabled', true);

```

 * `QUTE_VALIDATION_TEMPLATE_STATUS` is a Qute LS command server, so you define it in `QuteServerCommandConstants.QUTE_VALIDATION_TEMPLATE_STATUS` and remove `QuteSettings.QUTE_VALIDATION_TEMPLATE_STATUS`
 * you must implement `ConfigurationItemEditType.UpdateByRemoving` in order to remove the excluded patterns coming from the command which are declared in the `qute.validation.excluded` settings.json.

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +82,40 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
+      value: templateUri,
+      editType: ConfigurationItemEditType.Update,
+      section: QuteSettings.QUTE_VALIDATION_TEMPLATE_STATUS
+    } as ConfigurationItemEdit;
+    await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    await commands.executeCommand('setContext', 'quteValidationEnabled', !(result as any).validationEnabled);
+  }));
+  // Toggle off
+  context.subscriptions.push(commands.registerCommand(QuteSettings.QUTE_VALIDATION_ENABLED + QuteSettings.TOGGLE_OFF, async (uri?: Uri) => {
+    const templateUri = uri.toString();
```
angelozerr, line 119: In this case you don't need to consume the command, it shoud look like this:

```ts
// Add the template uri in the qute.validation.excluded settings
    const templateUri = uri.toString();
    const edit = {
      scopeUri: templateUri,
      value: templateUri,
      editType: ConfigurationItemEditType.Update,
      section: QuteSettings.QUTE_VALIDATION_EXCLUDED
    } as ConfigurationItemEdit;
    await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
    await commands.executeCommand('setContext', 'quteValidationEnabled', false);
```

angelozerr (changes requested):

src/extension.ts
```diff
@@ -51,8 +52,12 @@ export async function activate(context: ExtensionContext) {
   });
   // When a text document is opened,  update their language ID.
   context.subscriptions.push(
-    workspace.onDidOpenTextDocument((document) => {
-      updateLanguageId(document, true);
+    workspace.onDidOpenTextDocument(async (document) => {
+      if (!document.fileName.includes('.git')) {
```
rgrunber: You should have access to the full document Uri from `document`. The `scheme` of the Uri should just be `git`. This may be a bit more exact than checking the filename like this.

src/qute/commands/registerCommands.ts
AlexXuChen: This should be resolved now

package.json
```diff
@@ -159,6 +159,16 @@
         "command": "quarkusTools.deployToOpenShift.short",
         "title": "Deploy to OpenShift",
         "icon": "icons/openshift.png"
+      },
+      {
+        "command": "qute.validation.enabled.toggle.on",
```
rgrunber: So when you open a file, and validation is on, the command you run is toggle on ? Maybe it should be renamed to toggle off, so it makes more sense. It's really easy to get confused on what something does when you redefine terminology like this.

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +70,56 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
+      // Those patterns must be removed
+      const edit = {
+        scopeUri: templateUri,
+        value: result.excluded,
+        editType: ConfigurationItemEditType.Remove,
+        section: QuteSettings.QUTE_VALIDATION_EXCLUDED
+      } as ConfigurationItemEdit;
+      await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    }
+    await commands.executeCommand('setContext', 'quteValidationEnabled', true);
```
rgrunber: So you have a file opened, and validation is on. You click the validation icon, the language server confirms it's enabled, and the exclusion list is empty. So none of the above code paths are taken, and then you execute this line, which just enables validation, so nothing happens.

rgrunber (commented):
I would review the entire logic as a few things seem reversed.

package.json
AlexXuChen: I did it by naming convention since there are 2 separate buttons; if the open eye is shown, and the button is hovered, it displays that Qute validation is enabled for the current file. The inverse effect happens when clicking it (which is confusing implementation wise) where if you click the toggle on button, the button will switch to the eye closed button, but the toggle off effect will occur. The reverse happens for the toggle off state. I'm open to changing this however.

rgrunber: I just saw that you called the function `toggle.on`, and in it, you set `quteValidationEnabled` to `true`. Maybe if the function would have been called `toggle.off` it would have immediately made sense what the result should be.

AlexXuChen: Should the title be changed as well?

rgrunber: I would say yes because you're using the title to indicate the current state, but that should (hopefully) be achieved by the icon. It's common for the icon's hover text (title) to indicate what happens when you press it (eg. Enable Qute Validation)

src/extension.ts
```diff
@@ -49,10 +49,12 @@ export async function activate(context: ExtensionContext) {
   workspace.textDocuments.forEach(document => {
     updateLanguageId(document, false);
   });
-  // When a text document is opened,  update their language ID.
+  // When a text document is opened, update their language ID and check if qute is supported.
```
angelozerr: Remove qute comment

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +70,55 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
+ *
+ * @param context
+ */
+function registerQuteValidationToggleCommand(context: ExtensionContext) {
+  // Toggle off - add file to exclusion
+  context.subscriptions.push(commands.registerCommand(QuteSettings.QUTE_VALIDATION_ENABLED + QuteSettings.TOGGLE_OFF, async (uri?: Uri) => {
+    const templateUri = uri.toString();
+    // from QuteTemplateValidationStatusCommandHandler
+    const result: TemplateValidationStatus = await commands.executeCommand(QuteServerCommandConstants.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
+    if (result.validationEnabled) {
```
angelozerr: Why you check that?

package.json
```diff
@@ -257,6 +267,16 @@
           "command": "quarkusTools.deployToOpenShift.short",
           "when": "quarkusProjectExistsOrLightWeight && resourceExtname == .java",
           "group": "navigation@10"
+        },
+        {
+          "command": "qute.validation.enabled.toggle.off",
+          "when": "quarkusProjectExistsOrLightWeight && editorLangIdSupportsQute && quteValidationEnabled",
```
angelozerr: I  will rename quteValidationEnabled with editorQuteValidationEnabled

@rgrunber what do you think?

angelozerr (changes requested):

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +70,55 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
+      await commands.executeCommand(QuteClientCommandConstants.COMMAND_CONFIGURATION_UPDATE, edit);
+    }
+    await commands.executeCommand('setContext', 'quteValidationEnabled', false);
+  }));
+
+  // Toggle on - remove file(s) from exclusion
+  context.subscriptions.push(commands.registerCommand(QuteSettings.QUTE_VALIDATION_ENABLED + QuteSettings.TOGGLE_ON, async (uri?: Uri) => {
+    const templateUri = uri.toString();
+    const result: TemplateValidationStatus = await commands.executeCommand(QuteServerCommandConstants.QUTE_VALIDATION_TEMPLATE_STATUS, templateUri);
+    if (!result.validationEnabled) {
```
angelozerr, line 123: Please add à commenté which explains the update of settings

src/qute/commands/registerCommands.ts
```diff
@@ -69,6 +70,55 @@ function registerJavaDefinitionCommand(context: ExtensionContext) {
...
 }
 
+/**
+ * Toggle for enabling/disabling Qute validation.
+ *
+ * @param context
+ */
+function registerQuteValidationToggleCommand(context: ExtensionContext) {
+  // Toggle off - add file to exclusion
+  context.subscriptions.push(commands.registerCommand(QuteSettings.QUTE_VALIDATION_ENABLED + QuteSettings.TOGGLE_OFF, async (uri?: Uri) => {
```
angelozerr: Remove the constant with toggle off and on in settings and create 2
 constants in QuteClientCommandConstant.QUTE_VALIDATION_ENABLED_TOGGLE_ON and OFF

package.json
AlexXuChen: quteValidationEnabled(i.e. qute.validation.enabled) is the global setting, not the file specific setting, I feel editor in the name is unnecessary

angelozerr: quteValidationEnabled is the validation state for a given qute template (for a given editor), like editorLangIdSupportsQute,so I think we should start this property with `editor` prefix

src/qute/commands/commandConstants.ts
```diff
@@ -25,6 +25,16 @@ export namespace QuteClientCommandConstants {
    */
   export const COMMAND_CONFIGURATION_UPDATE = 'qute.command.configuration.update';
 
+  /**
+   * Toggle on extension for Qute validation setting.
+   */
+  export const TOGGLE_ON = '.toggle.on';
```
angelozerr: remove this constant

src/qute/commands/commandConstants.ts
```diff
@@ -25,6 +25,16 @@ export namespace QuteClientCommandConstants {
...
+
+  /**
+   * Toggle off extension for Qute validation setting.
+   */
+  export const TOGGLE_OFF = '.toggle.off';
```
angelozerr: remove this constant

src/qute/commands/commandConstants.ts
AlexXuChen: This was added based on your comment: https://github.com/redhat-developer/vscode-quarkus/pull/455#discussion_r803004245, would the full name be better here?

src/qute/commands/commandConstants.ts
AlexXuChen: same as above

src/qute/languageServer/client.ts
```diff
@@ -76,11 +77,22 @@ export function connectToQuteLS(context: ExtensionContext) {
       bindQuteRequest('qute/java/diagnostics');
       bindQuteRequest('qute/java/documentLink');
       bindQuteNotification('qute/dataModelChanged');
+
+      // When an editor is changed, check the qute validation context.
+      context.subscriptions.push(
+        window.onDidChangeActiveTextEditor(async editor => {
+          await updateQuteContext(editor.document);
```
rgrunber: I noticed when closing the last opened file, the active text editor is set to null, and this callback does get informed. This produces :

```
stack trace: TypeError: Cannot read property 'document' of undefined
	at /home/rgrunber/git/vscode-quarkus/dist/extension.js:85792:72
 ...
 ...
```

I don't think it affects anything but it does generate the error.

**Update** : It happens even when switching editors.

rgrunber:
I'm also seeing `command 'qute.command.validation.template.status' not found` if I have a file opened prior to the LS fully starting up. If validation should be on, it is respected but the icon shows validation as off.

If we don't have a way to fix this in time then let's just avoid displaying the error.

** Update ** : I'm guessing the client hasn't completed registration of the server-side command, even after `onready` ? Is there a way to fix this @angelozerr ?

src/qute/commands/registerCommands.ts
angelozerr: Yes the full name without concat

src/qute/commands/commandConstants.ts
angelozerr: Yes please

src/qute/languageServer/client.ts
angelozerr: I think @AlexXuChen have fixed that. If it's true @AlexXuChen could you `Resolve conversation` please.

src/qute/commands/registerCommands.ts
```diff
@@ -91,17 +141,44 @@ export function registerConfigurationUpdateCommand(context: ExtensionContext) {
...
     }
   }));
 }
 
-function getSettingsValue(value: any, section: string) {
+/**
+   * Sets the `editorLangIdSupportsQute` context to `true` if the editor language id
+   * is a qute template. Sets to `false` otherwise.
+   */
+ export async function updateQuteContext(document: TextDocument) {
```
angelozerr: Format this function

src/qute/commands/registerCommands.ts
```diff
@@ -91,17 +165,44 @@ export function registerConfigurationUpdateCommand(context: ExtensionContext) {
...
+}
+
+function getSettingsValue(value: any, section: string, editType: ConfigurationItemEditType) {
   switch (section) {
     case QuteSettings.QUTE_VALIDATION_EXCLUDED:
-      return workspace.asRelativePath(Uri.parse(value), false);
-    default:
-      return value;
+      if (editType === ConfigurationItemEditType.Add) {
+        return workspace.asRelativePath(Uri.parse(value), false);
```
AlexXuChen, line 210: This logic seems to add an extra path separator(e.g. `/src...` instead of `src...` ) at the beginning on occasion, and so I feel this settings value retrieval introduces more problems then it fixes.

angelozerr: In Windows OS I have not this sproblem.

@fbricon @rgrunber do you encounter this problem?

angelozerr: @AlexXuChen this code is used to have relative pattern from workspace folder instead of having file:// uri pattern.

@AlexXuChen what is your suggestion?

angelozerr:
> I'm also seeing command 'qute.command.validation.template.status' not found if I have a file opened prior to the LS fully starting up. 

This issue should be fixed now?

src/qute/commands/registerCommands.ts
AlexXuChen: Using workspace.asRelativePath(Uri.parse(value), false) seems to add more problems than just using the raw template uri, since the template uri is already relative to the workspace. I suggest we dont use getSettingsValue simply because it introduces this problem.

AlexXuChen: I believe we can achieve the same result by calling path.relative(workspace.workspaceFolders[0].uri.toString(), window.activeTextEditor.document.uri.toString()) since we're looking at the activeTextEditor and this is a workplace setting

rgrunber:
> > I'm also seeing command 'qute.command.validation.template.status' not found if I have a file opened prior to the LS fully starting up.
> 
> This issue should be fixed now?

It's fixed now.

I see the following :

1) Keep a Qute file opened, and disable the validation
2) Reload the window
3) The validation will show as enabled once everything has loaded even though the file is definitely excluded

src/qute/commands/registerCommands.ts
angelozerr: As I said this code avoids storing template uri with file:// scheme which is not portable. Try to remove getseetingsvalue and you will see what I mean.

Im surprised that you have some trouble (which trouble?) because this code has been tested before by @fbricon and me and it works good.

Could you explain what is your problem please?

AlexXuChen:
> I see the following :
> 
>     1. Keep a Qute file opened, and disable the validation
> 
>     2. Reload the window
> 
>     3. The validation will show as enabled once everything has loaded even though the file is definitely excluded

I've noticed the validation displays as enabled after the Quarkus support starts even though the file is in the exclusion. It seems the https://github.com/redhat-developer/vscode-quarkus/pull/455/files#diff-c1b7030f8c5f91b6d8690295e037104c11aecfc6d7eb7a9bd531aa0a9c5547fdR91 check doesn't handle this case, and we might have to set the context on extension `activate`

rgrunber:
What if you called `setQuteValidationEnabledContext()` before the active editor checks ?

**Update** : I think it works. The global context gets set first, and if an opened file with exlcudes should override something, it gets handled afterwards. This falls in line with our idea of local behaviour overriding global.

AlexXuChen:
> What if you called `setQuteValidationEnabledContext()` before the active editor checks ?

Moving to before the activeTextEditor as opposed to after which it was originally seems to fix the issue

src/qute/commands/registerCommands.ts
AlexXuChen: The toggle command is visible from the command palette, however, when we trigger that command, the uri doesnt exist. To fix this, I added an undefined check to use the path relative to the editor file and the workspace folder. However, when this uri is appended to the configuration edit, getSettingsValue adds a leading `/`. Can we avoid this and just find the relative path in the command listener instead?

AlexXuChen: i.e. my fix in toggle on/off:
```
    const templateUri = uri ? uri.toString() : path.relative(workspace.workspaceFolders[0].uri.toString(), window.activeTextEditor.document.uri.toString());

```

rgrunber:
@angelozerr I'm seeing the following stacktrace when I close a file such as `/home/rgrunber/git/quarkus-ls/qute.ls/com.redhat.qute.ls/src/test/resources/templates/base.qute.html` :

Looks like the template directory is set as under `src/main` for the project but that doesn't exist in this case.

<details>
<summary>stacktrace</summary>

```
java.nio.file.NoSuchFileException: /home/rgrunber/git/quarkus-ls/qute.ls/com.redhat.qute.ls/src/main/resources/templates
    at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
    at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
    at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
    at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
    at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
    at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
    at java.base/java.nio.file.Files.readAttributes(Files.java:1764)
    at java.base/java.nio.file.FileTreeWalker.getAttributes(FileTreeWalker.java:219)
    at java.base/java.nio.file.FileTreeWalker.visit(FileTreeWalker.java:276)
    at java.base/java.nio.file.FileTreeWalker.walk(FileTreeWalker.java:322)
    at java.base/java.nio.file.FileTreeIterator.<init>(FileTreeIterator.java:71)
    at java.base/java.nio.file.Files.walk(Files.java:3825)
    at java.base/java.nio.file.Files.walk(Files.java:3879)
    at com.redhat.qute.project.indexing.QuteIndexer.scan(QuteIndexer.java:84)
    at com.redhat.qute.project.indexing.QuteIndexer.lambda$scanAsync$0(QuteIndexer.java:74)
    at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
    at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692)
    at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
    at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
    at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
    at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
    at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
```
</details>

**Update** : Aha https://github.com/redhat-developer/quarkus-ls/blob/master/qute.jdt/com.redhat.qute.jdt/src/main/java/com/redhat/qute/jdt/utils/JDTQuteProjectUtils.java#L42 . So likely that [`project.getTemplateBaseDir()`](https://github.com/redhat-developer/quarkus-ls/blob/master/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/project/indexing/QuteIndexer.java#L84) just needs a check for existence.

src/qute/commands/registerCommands.ts
angelozerr: Please keep the existing coďe. Your code doesnt tzke care of multiple workspace folders.

Your problem is when you execute command with palette if I understand.

To fix that you should take the document uri of the current text editor.

angelozerr:
@rgrunber as you noticed its a bug of qute ls which is not related to this issue. Please create the issue in quarkus ls.Thanks

src/qute/commands/registerCommands.ts
angelozerr: Newbie question: can we avoid having those commands in  ommand palette?

rgrunber: Did you try the most recent PR ? I don't see the commands in the command palette. The command palette contribution of the commands with `"when": "false"` seems to work.

angelozerr: Nice! @AlexXuChen is it good now to keep the existing code with workspace.asRelativePath(Uri.parse(value), false) ? Or do you have another usecase which causes problems ?

angelozerr:
@rgrunber @AlexXuChen if I understand correctly this PR should work now?

AlexXuChen:
> @rgrunber @AlexXuChen if I understand correctly this PR should work now?

If we're okay to hid the enable/disable from command palette, then yes the issue is finished

src/qute/commands/registerCommands.ts
AlexXuChen: If the user can't use the command palette commands, then yser everything is working as intended

fbricon:
I'm fine hiding the command from the palette, but my main problem is the toggle icon doesn't appear when I reopen vscode:

https://user-images.githubusercontent.com/148698/153577662-c8eb400d-0293-45d7-bf18-eb9479c55fa1.mov

fbricon:
Same thing happens when opening a new file. Potentially, this is caused by the button's activation clause being computed before the mechanism to force  the qute language runs. That's why revisiting the same open file works.

Now, if you have the toggle button that shows the validation is on, and you apply a quickfix to disable validation for that particular file, from the problems view, the button state is not updated.

Feb-11-2022 11-56-04

angelozerr:
@fbricon could you try my POC https://github.com/redhat-developer/vscode-quarkus/pull/462 ?

My fix is to track the change of language id and refresh the Qute context in this case https://github.com/redhat-developer/vscode-quarkus/pull/462/commits/995ea7bedd0e8684ba57bd5fc50e5a69abbef8b3#diff-c1b7030f8c5f91b6d8690295e037104c11aecfc6d7eb7a9bd531aa0a9c5547fdR93

I use onDidOpenTextDocument because the doc onDidOpenTextDocument: [Event](https://code.visualstudio.com/api/references/vscode-api#Event)<[TextDocument](https://code.visualstudio.com/api/references/vscode-api#TextDocument)>
 says:

```
An event that is emitted when a [text document](https://code.visualstudio.com/api/references/vscode-api#TextDocument) is opened or when the language id of a text document [has been changed](https://code.visualstudio.com/api/references/vscode-api#languages.setTextDocumentLanguage).

To add an event listener when a visible text document is opened, use the [TextEditor](https://code.visualstudio.com/api/references/vscode-api#TextEditor) events in the [window](https://code.visualstudio.com/api/references/vscode-api#window) namespace. Note that:

The event is emitted before the [document](https://code.visualstudio.com/api/references/vscode-api#TextDocument) is updated in the [active text editor](https://code.visualstudio.com/api/references/vscode-api#window.activeTextEditor)
When a [text document](https://code.visualstudio.com/api/references/vscode-api#TextDocument) is already open (e.g.: open in another [visible text editor](https://code.visualstudio.com/api/references/vscode-api#window.visibleTextEditors)) this event is not emitted
```

The problem is that we update qute context twice now when we open a qute template file (one by onDidOpenTextDocument 
 at first followed by onDidChangeActiveTextEditor)

angelozerr:
In some case (I need to retrieve it), my button disappears after click on it.

src/qute/commands/registerCommands.ts
```diff
@@ -91,17 +165,65 @@ export function registerConfigurationUpdateCommand(context: ExtensionContext) {
...
     }
   }));
 }
 
-function getSettingsValue(value: any, section: string) {
+/**
+   * Sets the `editorLangIdSupportsQute` context to `true` if the editor language id
+   * is a qute template. Sets to `false` otherwise.
+   */
+export async function updateQuteContext(document: TextDocument) {
```
rgrunber: What's the difference between this and `synchronizeQuteValidationButton` ? With the new change this isn't used anywhere. Can it be removed ?

angelozerr: The merge from my POC was not done correctly,now we have synchronizeQuteValidationButton instead of updateQuteContext.

angelozerr: updateQuteContext must be removed

angelozerr:
@fbricon it should be good now (I hope), please retry it.

rgrunber:
There's still one more issue I'm seeing in a Quarkus project when `quarkus.tools.propertiesLanguageMismatch` is `prompt`. When the file (eg. hello.html) is opened, the prompt correctly comes up via. `onDidOpenTextDocument` and the language gets updated correctly. Once the language gets updated, `onDidOpenTextDocument` is triggered yet again (just as the API says it should). I don't see any code to check that the `languageId` of the document is already set to something `qute-*` or something like that so the prompt comes up again. This time the `languageId` isn't changed, so the prompt doesn't come up a 3rd time.

Is it enough to just check whether the `languageId` is `qute-*` much like we check in `updateLanguageId` for `quarkus-properties` ?

angelozerr:
@rgrunber I agree with you there is à problem  with popup which appears twice or appear altough the file has already the qute language.

But this problem comes from master too, no?

If it that those problems should be fixed in 1n another PR.

If it is not that have you a usecase which causes problem?

rgrunber:
If it exists already outside this PR then let's leave it for now. The PR fixed the previous issue.

fbricon:
Thanks @AlexXuChen . Let's fix the double popup issue with #459

PR #456, opened 2022-02-02
POC button

angelozerr:
POC button (please don't merge it)

### DIFF
--- a/src/qute/commands/registerCommands.ts
+++ b/src/qute/commands/registerCommands.ts
@@ -14,6 +14,22 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
   registerGenerateTemplateFileCommand(context);
   registerJavaDefinitionCommand(context);
   registerConfigurationUpdateCommand(context);
+  registerVal(context);
+}
+
+function registerVal(context: ExtensionContext) {
+  context.subscriptions.push(commands.registerCommand('qute.validation.enabled.toggle', async (uri?: Uri) => {
+    const templateUri =   uri.toString();
+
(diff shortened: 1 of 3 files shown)


src/qute/commands/registerCommands.ts
```diff
@@ -14,6 +14,22 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
...
   registerJavaDefinitionCommand(context);
   registerConfigurationUpdateCommand(context);
+  registerVal(context);
+}
+
+function registerVal(context: ExtensionContext) {
+  context.subscriptions.push(commands.registerCommand('qute.validation.enabled.toggle', async (uri?: Uri) => {
+    const templateUri =   uri.toString();
+
+    const result = await commands.executeCommand('qute.command.validation.template.status', templateUri);
```
AlexXuChen, line 24: It seems this command doesn't exist. Is this meant to check if  the templateUri exists in the qute.validation.excluded array?

angelozerr: This command exists when qute ls is started. This command gives you information about why the qute template validation is enabled or not.

A qute template validation is enabled if qute.validation.enabled is true and the templateuri doesnt matches one or several patterns from qute.validation.excluded.

A qute template validation is disabled if qute.validation.enabled is set to false or templateuri matches one or several patterns from qute.validation.excluded.

In other words according to the command result you can know if the qute template must be validated or not AND which settings you must update to enable or disable the validation.

angelozerr:
Close for https://github.com/redhat-developer/vscode-quarkus/pull/455

PR #459, opened 2022-02-09
Added `qute.templates.languageMismatch` to avoid reuse of `quarkus.tools.propertiesLanguageMismatch`

### LINKED ISSUE #457: Don't  reuse quarkus.tools.propertiesLanguageMismatch for Qute
When we force the languageId for Qute templates, we should not reuse `"quarkus.tools.propertiesLanguageMismatch": "forceQuarkus",`, but have a dedicated `"qute.templates.languageMismatch": "forceQute",`

AlexXuChen:
Added `qute.templates.languageMismatch` to avoid reuse of `quarkus.tools.propertiesLanguageMismatch`.

All pop-ups and setting configurations are now Qute specific:
Screenshot from 2022-02-09 15-53-54
Screenshot from 2022-02-09 15-54-05
Screenshot from 2022-02-09 15-54-36

Fixes #457 

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,16 +207,32 @@
           "type": "string",
           "enum": [
             "ignore",
-            "forceQuarkus",
+            "force",
             "prompt"
           ],
           "enumDescriptions": [
-            "ignore language mismatch for Quarkus properties",
+            "Ignore language mismatch for Quarkus properties",
             "Automatically correct the language for mismatched Quarkus properties",
             "Prompt correcting the language for mismatched Quarkus properties"
           ],
-          "default": "forceQuarkus",
+          "default": "force",
           "description": "Action performed when detected Quarkus properties have an incorrect language.",
+          "scope": "window"
+        },
+        "qute.templates.languageMismatch": {
+          "type": "string",
+          "enum": [
+            "ignore",
+            "force",
+            "prompt"
+          ],
+          "enumDescriptions": [
+            "Ignore language mismatch for Qute templates",
+            "Automatically correct the language for mismatched Qute templates",
+            "Prompt correcting the language for mismatched Qute templates"
+          ],
+          "default": "force",
+          "description": "Action performed when detected Qute templates have an incorrect language.",
           "scope": "window"
         },
         "qute.trace.server": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -17,11 +17,12 @@ import * as path from 'path';
 import { commands, ConfigurationChangeEvent, Disposable, ExtensionContext, extensions, languages, Terminal, TextDocument, window, workspace } from 'vscode';
 import { registerVSCodeCommands } from './commands/registerCommands';
 import { ProjectLabelInfo } from './definitions/ProjectLabelInfo';
-import { PropertiesLanguageMismatch, QuarkusConfig } from './QuarkusConfig';
+import { QuarkusPropertiesLanguageMismatch, QuarkusConfig } from './QuarkusConfig';
 import { QuarkusContext } from './QuarkusContext';
 import quarkusProjectListener from './QuarkusProjectListener';
 import { connectToQuteLS } from './qute/languageServer/client';
 import { terminalCommandRunner } from './terminal/terminalCommandRunner';
+import { tryToForceLanguageId } from './utils/languageMismatch';
 import { JAVA_EXTENSION_ID } from './utils/requestStandardMode';
 import { initTelemetryService } from './utils/telemetryUtils';
 import { WelcomeWebview } from './webviews/WelcomeWebview';
@@ -51,13 +52,13 @@ export async function activate(context: ExtensionContext) {
 
   // When extension is started, loop for each text documents which are opened to update their language ID.
   workspace.textDocuments.forEach(document => {
-    updateLanguageId(document, false);
+    updateLanguageId(context, document, false);
   });
   // When a text document is opened, update their language ID.
   context.subscriptions.push(
     workspace.onDidOpenTextDocument(async (document) => {
       if (!(document.uri.scheme === 'git')) {
-        await updateLanguageId(document, true);
+        await updateLanguageId(context, document, true);
       }
     })
   );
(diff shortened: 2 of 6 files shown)


src/qute/commands/registerCommands.ts
```diff
@@ -104,6 +112,92 @@ function getSettingsValue(value: any, section: string) {
   }
 }
 
+const LANGUAGE_MAP = new Map<string, string>([
+  [".html", "qute-html"],
```
fbricon, line 231: add .htm and .yml

package.json
```diff
@@ -209,6 +209,22 @@
...
         },
+        "qute.templates.languageMismatch": {
+          "type": "string",
+          "enum": [
+            "ignore",
+            "forceQute",
+            "prompt"
+          ],
+          "enumDescriptions": [
+            "ignore language mismatch for Qute templates",
```
fbricon: Ignore

fbricon:
still draft?

AlexXuChen:
> still draft?

We wanted to move the shared functionality between Qute and Quarkus to a shared utils file. If you think it's better that we do that in a separate pr, I can create a separate issue and merge this one.

AlexXuChen:
There is also this existing issue related to `quarkus.tools.propertiesLanguageMismatch`: https://github.com/redhat-developer/vscode-quarkus/pull/455#issuecomment-1036734047

src/qute/commands/registerCommands.ts
```diff
@@ -104,6 +112,94 @@ function getSettingsValue(value: any, section: string) {
...
+  // Get project label information for the given file URI
+  const labelInfo = ProjectLabelInfo.getProjectLabelInfo(document.uri.toString());
+  return labelInfo.then(l => {
+    if (l.isQuarkusProject()) {
+      if (propertiesLanguageMismatch === QuteTemplateLanguageMismatch.prompt) {
+        const YES = "Yes", NO = "No";
+        const response: Thenable<string> = window.showInformationMessage(`The language ID for '${fileName}' must be '${languageId}' to receive Qute support. Set the language ID to '${languageId}?'`, YES, NO);
+        response.then(result => {
+          if (result === YES) {
+            // The application.properties file belong to a Quarkus project, force to the quarkus-properties language
```
fbricon: not an application.properties

src/qute/commands/registerCommands.ts
```diff
@@ -104,6 +112,94 @@ function getSettingsValue(value: any, section: string) {
...
+            languages.setTextDocumentLanguage(document, languageId);
+          }
+        });
+      } else {
+        // The application.properties file belong to a Quarkus project, force to the quarkus-properties language
```
fbricon: not an application.properties

src/qute/commands/registerCommands.ts
```diff
@@ -104,6 +112,94 @@ function getSettingsValue(value: any, section: string) {
...
+
+/**
+ * Update the language ID to a supported Qute language id if needed.
+ *
+ * @param document the text document.
+ * @param onExtensionLoad if the user manually changed the language id.
+ */
+ async function updateQuteLanguageId(context: ExtensionContext, document: TextDocument, onExtensionLoad: boolean) {
+  const propertiesLanguageMismatch: QuteTemplateLanguageMismatch = QuteSettings.getQuteTemplatesLanguageMismatch();
+  if (propertiesLanguageMismatch === QuteTemplateLanguageMismatch.ignore) {
```
fbricon: || document.languageId.startsWith('qute-') // the language ID is already qute-* do nothing.

to fix https://github.com/redhat-developer/vscode-quarkus/pull/455#issuecomment-1036734047

src/utils/languageMismatch.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  static saveToConfig<T>(configName: string, value: T): Thenable<void> {
+    return workspace.getConfiguration().update(configName, value, ConfigurationTarget.Global);
+  }
+}
+
+enum LanguageMismatchOption {
+  ignore = "ignore",
+  force = "force",
```
AlexXuChen: Renamed option to `force` for consistency in both Qute and Quarkus language mismatch

package.json
```diff
@@ -207,18 +207,34 @@
           "type": "string",
           "enum": [
             "ignore",
-            "forceQuarkus",
```
fbricon, line 210: what happens if you already had `forceQuarkus` set?

fbricon: well the good news is unknown values are considered `force`, since the code doesn't check it, so, no regression there; the bad news is unknown values are considered `force`, since the code doesn't check it, so, not great.

fbricon:
So looks like when we reopen vscode, if a qute template was opened previously, it no longer is forced to a qute-* language on startup.

AlexXuChen:
> So looks like when we reopen vscode, if a qute template was opened previously, it no longer is forced to a qute-* language on startup.

Is this a file with an extension without `qute-` (i.e. `a.html` not `a.qute-html`)? It's possible that this issue is existing, since the non-qute extensions aren't included in the activation event: https://github.com/redhat-developer/vscode-quarkus/blob/44469274c7d09b3795485e563550de8116b8c733/package.json#L36-L51 because I believe I've seen similar behaviour before.

fbricon:
I double-checked, it works properly in master. If you had an opened src/main/resources/templates/foo.html before restart, on restart, once the quarkus extension is activated, the languageId is automatically switched to qute-html

package.json
AlexXuChen: Could you be a bit more specific? Are you saying that now it's attempting to force non-qute supported languages as well?

rgrunber:
> I double-checked, it works properly in master. If you had an opened src/main/resources/templates/foo.html before restart, on restart, once the quarkus extension is activated, the languageId is automatically switched to qute-html

I see the same thing with this change though. I have `qute.templates.languageMismatch` set to `force` and once I manually activate Quarkus (eg. open some Java file), the `languageId` for the template .html file always gets set to qute-html. Am I missing something here ? It looks like it's working correctly.

AlexXuChen:
> I double-checked, it works properly in master. If you had an opened src/main/resources/templates/foo.html before restart, on restart, once the quarkus extension is activated, the languageId is automatically switched to qute-html

I see the same thing as Roland, where this is working on this change; once the Quarkus extension has started. The thing is that the Quarkus extension doesn't start if the open editor is not a `qute-*` extension, so the force won't occur.

AlexXuChen:
However, I do see an issue on text editor change, where if the Quarkus extension starts and a qute supported file is open, switching to it won't force the languageId, but reopening it will. We just need to add a text editor change event.
updatelanguageid

src/qute/commands/registerCommands.ts
```diff
@@ -17,6 +19,11 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
   registerJavaDefinitionCommand(context);
   registerConfigurationUpdateCommand(context);
   registerQuteValidationToggleCommand(context);
+  context.subscriptions.push(
+    workspace.onDidOpenTextDocument((document) => {
+      updateQuteLanguageId(context, document, true);
```
rgrunber: I think it would be worth it to surround the update with :

```
if (!(document.uri.scheme === 'git')) {
...
}
```

like we do at https://github.com/redhat-developer/vscode-quarkus/blob/master/src/extension.ts#L59 .

src/qute/commands/registerCommands.ts
```diff
@@ -17,6 +19,11 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
   registerJavaDefinitionCommand(context);
   registerConfigurationUpdateCommand(context);
   registerQuteValidationToggleCommand(context);
+  context.subscriptions.push(
```
angelozerr, line 22: To fix the issue from @fbricon (when vscode starts with one HTML file opened), please add

```ts
// When extension is started, loop for each text documents which are opened to update their language ID.
  workspace.textDocuments.forEach(document => {
    updateQuteLanguageId(context, document, false);
  });
```

exactly like application.properties file.

fbricon: yup; see https://github.com/redhat-developer/vscode-quarkus/blob/44469274c7d09b3795485e563550de8116b8c733/src/extension.ts#L53-L55

src/extension.ts
```diff
@@ -84,80 +85,18 @@ function displayWelcomePageIfNeeded(context: ExtensionContext): void {
...
  * Update if required the language ID to 'quarkus-properties' if needed.
  *
+ * @param context the extension context
  * @param document the text document.
  * @param onExtensionLoad if the user manually changed the language id.
  */
-async function updateLanguageId(document: TextDocument, onExtensionLoad: boolean) {
-  const propertiesLanguageMismatch: PropertiesLanguageMismatch = QuarkusConfig.getPropertiesLanguageMismatch();
-  if (propertiesLanguageMismatch === PropertiesLanguageMismatch.ignore) {
+async function updateLanguageId(context: ExtensionContext, document: TextDocument, onExtensionLoad: boolean) {
```
fbricon, line 97: should also handle qute templates

src/qute/commands/registerCommands.ts
AlexXuChen: This does the same thing as an active text editor listener, so I believe we can leave out the text editor listener and go with this (since this is only an edge case on Quarkus extension start)

src/extension.ts
AlexXuChen: There is already support for Qute templates, I simply moved it to where we have all our Qute logic

src/qute/languageServer/settings.ts
```diff
@@ -17,4 +19,20 @@ export namespace QuteSettings {
...
+
+  /**
+   * Action performed when detected Qute templates have an incorrect language.
+   */
+  export const QUTE_TEMPLATES_LANGUAGE_MISMATCH = "qute.templates.languageMismatch";
+
+  /**
+   * Flag for enabling forcing Qute language id setting
+   */
+  export const QUTE_OVERRIDE_LANGUAGE_ID = 'qute.templates.override.languageId.ttttt';
```
rgrunber: This shouldn't be here right ?

AlexXuChen: Ah sorry I was testing the flag

AlexXuChen: Keep forgetting to change it back

src/qute/commands/registerCommands.ts
```diff
@@ -214,6 +227,42 @@ function getSettingsValue(value: any, section: string, editType: ConfigurationIt
...
+ */
+ async function updateQuteLanguageId(context: ExtensionContext, document: TextDocument, onExtensionLoad: boolean) {
+  const propertiesLanguageMismatch: QuteTemplateLanguageMismatch = QuteSettings.getQuteTemplatesLanguageMismatch();
+  // Check if the setting is set to ignore or if the language ID is already set to Qute
+  if (propertiesLanguageMismatch === QuteTemplateLanguageMismatch.ignore || document.languageId.startsWith('qute-')) {
+    // Do nothing
+    return;
+  }
+  const fileName: string = path.basename(document.fileName);
+  if (document.fileName.includes(`resources${path.sep}templates${path.sep}`)) {
```
angelozerr, line 254: When a HTML file comes from JAR, the document.fileName looks like this `\\templates\\tags\\ifError.html`, it should be really nice if you could support that too.

AlexXuChen: If you're referring to a supported file extension that is nested in a directory within `resources/templates`, this is already supported:
nestedtemplatelanguagesupport

angelozerr: No I speak about html file hosted in a JAR. Please ignore my comment we will fix that in an another PR

rgrunber:
I've tried out maintaining an open `.html` file and activating the extension. It's working correctly for me now. I think this is ready to merge.

fbricon:
Works fine now, thanks @AlexXuChen!

PR #461, opened 2022-02-10
Use embedded JRE from redhat.java to launch Qute language server.

rgrunber:
- Create helper method getJavaExtensionAPI so that it may be reused
- Use 'redhat.java' embedded JRE through RequirementsData from its
  ExtensionAPI
- Adjust README with new requirements


fbricon:
thanks @rgrunber

PR #474, opened 2022-03-11
Support for `textDocument/InlayHint`

angelozerr:
Support for `textDocument/InlayHint`

See https://github.com/redhat-developer/quarkus-ls/issues/595

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "bugs": "https://github.com/redhat-developer/vscode-quarkus/issues",
   "engines": {
-    "vscode": "^1.37.0"
+    "vscode": "^1.65.0"
   },
   "galleryBanner": {
     "color": "#d8ebff",
@@ -218,11 +218,6 @@
           "description": "Action performed when detected Quarkus properties have an incorrect language.",
           "scope": "window"
         },
-        "qute.codeLens.enabled": {
-          "type": "boolean",
-          "default": true,
-          "markdownDescription": "Enable/disable Qute CodeLens. Default is `true`."
-        },
         "qute.templates.languageMismatch": {
           "type": "string",
           "enum": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "bugs": "https://github.com/redhat-developer/vscode-quarkus/issues",
   "engines": {
-    "vscode": "^1.37.0"
+    "vscode": "^1.65.0"
```
fbricon, line 13: most likely will be incompatible with theia

fbricon: @benoitf what's the max vscode engine version compatible with theia?

angelozerr: > most likely will be incompatible with theia

please note vscode-java have a PR to support Inlay Hint too, and they did that too https://github.com/redhat-developer/vscode-java/pull/2354/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R26

rgrunber: Yeah, it's not supported currently. https://eclipse-theia.github.io/vscode-theia-comparator/status.html .

Also, from https://github.com/redhat-developer/vscode-java/pull/2183#pullrequestreview-790383636, looks like updating the engine itself should not be harmful to Theia. Just have to ensure you don't call API that doesn't exist.

**Update** : According to https://github.com/eclipse-theia/theia/blob/master/dev-packages/application-package/src/api.ts#L21, should be 1.53.2

rgrunber: Also see https://github.com/redhat-developer/vscode-java/pull/1975/files for how to adopt new API while still being compatible with older engines.

AlexXuChen (approved):
LGTM, setting works as intended

angelozerr:
Thanks @AlexXuChen for the review!

PR #479, opened 2022-03-24
Changed Qute pop-up to appear on text document open

AlexXuChen:
Changed Qute pop-up to appear on text document open to fix for UI test timeout

### DIFF
--- a/src/qute/languageServer/client.ts
+++ b/src/qute/languageServer/client.ts
@@ -100,14 +100,23 @@ export function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI
           if (window.activeTextEditor?.document === document) {
             await synchronizeQuteValidationButton(window.activeTextEditor);
           }
+          // Display the experimental Qute validation pop-up if it hasn't been displayed and a Qute file is open
+          if (!hasShownQuteValidationPopUp(context) && document.languageId.includes('qute')) {
+            showQuteValidationPopUp(context);
+          }
         })
       );
+      // Display the experimental Qute validation pop-up if it hasn't been displayed and a Qute file is open
+      if (!hasShownQuteValidationPopUp(context)) {
+        for (let textDocument of workspace.textDocuments) {
+          if (textDocument.languageId.includes('qute')) {
+            showQuteValidationPopUp(context);
+          }
+        }
+      }
       await setQuteValidationEnabledContext();
       await synchronizeQuteValidationButton(window.activeTextEditor);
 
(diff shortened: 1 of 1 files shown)


rgrunber:
If you rebase against the latest version, it should be good again.

AlexXuChen:
The tests still fail with my side, potentially we just stay with your fix then @rgrunber and I can just close this pr?

src/qute/languageServer/client.ts
```diff
@@ -104,9 +104,12 @@ export function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI
       await setQuteValidationEnabledContext();
       await synchronizeQuteValidationButton(window.activeTextEditor);
 
-      if (!hasShownQuteValidationPopUp(context)) {
-        showQuteValidationPopUp(context);
-      }
+      context.subscriptions.push(
+        workspace.onDidOpenTextDocument((document) => {
+          if (!hasShownQuteValidationPopUp(context)) {
+            showQuteValidationPopUp(context);
```
rgrunber: You could add this line to the same invocation from a few lines above.

rgrunber (commented):
With this change, if I simple activate the Quarkus extension by going to the welcome page, and then open an independent XML file, the experimental validation popup comes up. That shouldn't happen.

The purpose of this PR would be to only bring up the popup if the opened file is a Qute template, so there should be some code that checks whether that's the case.

src/qute/languageServer/client.ts
```diff
@@ -100,14 +100,15 @@ export function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI
           if (window.activeTextEditor?.document === document) {
             await synchronizeQuteValidationButton(window.activeTextEditor);
           }
+          // Display the experimental Qute validation pop-up if it hasn't been displayed
+          if (!hasShownQuteValidationPopUp(context)) {
```
rgrunber: Can we do something like :

```
  const quteSupported = document.languageId.includes('qute');
  if (quteSupported) {
    showQuteValidationPopUp(context);
  }
```
Without this, the popup would come up for even Java source files.

AlexXuChen: It seems both checks on gh actions have passed finally?
image
Don't know if this will be consistent.

rgrunber: Could just be luck of the draw. Either way, would be good to investigate those tests eventually.

rgrunber (approved):

PR #480, opened 2022-03-24
Improved syntax coloration for operator support for #if

### LINKED ISSUE #478: Syntax coloration for operator support for #if
The operators used in #if must be improved. At first the operator must be available only for #if. Here a sample with `eq`:

image

The `eq` used in #if is higlighted as purple (it's correct), but it is higlighted too inside expression (the second eq), which is false.

It misses some operator like `and`. You can find all alowed operator for #if at https://quarkus.io/guides/qute-reference#if_section

You can see too Junit at https://github.com/quarkusio/quarkus/blob/main/independent-projects/qute/core/src/test/java/io/quarkus/qute/IfSectionTest.java with #if  to have some sample

AlexXuChen:
Improved syntax colouration for operator support for #if based on https://quarkus.io/guides/qute-reference#if_section, and separated by precedence.

Fixes #478 

### DIFF
--- a/language-support/qute/qute.tmLanguage.json
+++ b/language-support/qute/qute.tmLanguage.json
@@ -21,6 +21,12 @@
     "templates": {
       "patterns": [
         {
+          "include": "#section_start_if_tag"
+        },
+        {
+          "include": "#section_start_for_tag"
+        },
+        {
           "include": "#section_start_default_tag"
         },
         {
@@ -86,7 +92,7 @@
 			"name": "string.unquoted.cdata.qute"
 		},
     "section_start_default_tag": {
-      "begin": "({)(#)((each|else|else\\sif|eval|for|if|include|insert|set|let|with|switch|case|is|when)\\b)",
+      "begin": "({)(#)((each|else|else\\sif|eval|include|insert|set|let|with|switch|case|is|when)\\b)",
       "end": "(\\/)?((?<![\\\\])})",
       "beginCaptures": {
         "1": {
@@ -132,6 +138,74 @@
           "name": "support.constant.handlebars"
         }
       }
+    },
+    "section_start_if_tag": {
+      "begin": "({)(#)((if)\\b)",
+      "end": "(\\/)?((?<![\\\\])})",
+      "beginCaptures": {
+        "1": {
+          "name": "support.constant.handlebars"
+        },
+        "2": {
+          "name": "keyword.control"
+        },
+        "3": {
+          "name": "keyword.control"
+        }
+      },
+      "endCaptures": {
+        "1": {
+          "name": "keyword.control"
+        },
+        "2": {
+          "name": "support.constant.handlebars"
+        }
+      },
+      "patterns": [
+        {
+          "match": "\\b(gt|ge|lt|le|eq|is|ne|and|or)\\b",
+          "name": "keyword.control.qute"
+        },
+        {
+          "match": "\\b(!|>|>=|<|<=|==|!=|&&|\\|\\|)\\b",
+          "name": "keyword.operator.qute"
+        },
+        {
+          "include": "#code"
+        }
+      ]
+    },
+    "section_start_for_tag": {
+      "begin": "({)(#)((for)\\b)",
+      "end": "(\\/)?((?<![\\\\])})",
+      "beginCaptures": {
+        "1": {
+          "name": "support.constant.handlebars"
+        },
+        "2": {
+          "name": "keyword.control"
+        },
+        "3": {
+          "name": "keyword.control"
+        }
+      },
+      "endCaptures": {
+        "1": {
+          "name": "keyword.control"
+        },
+        "2": {
+          "name": "support.constant.handlebars"
+        }
+      },
+      "patterns": [
+        {
+          "match": "\\bin\\b",
+          "name": "keyword.control.flow"
+        },
+        {
+          "include": "#code"
+        }
+      ]
     },
     "section_start_user_tag": {
       "begin": "({)(#)(\\w+(\\.\\w+)*)",
@@ -199,13 +273,17 @@
       },
       "patterns": [
         {
+          "match": "\\b(or|orEmpty)\\b",
+          "name": "keyword.control.qute"
+        },
+        {
+          "match": "\\b(&&|\\|\\||\\?:|\\?|:)\\b",
+          "name": "keyword.operator.qute"
+        },
+        {
           "include": "#code"
         }
       ]
-    },
-    "expression_content": {
-      "match": "[a-zA-Z_\\(][\\w\\(\\)+-\/\\*]+(\\.?[\\w\\(\\)+-\/\\*]+)*",
-      "name": "variable.other.readwrite.local.qute"
     },
     "code": {
       "patterns": [
@@ -220,9 +298,6 @@
         },
         {
           "include": "#lambda-expression"
-        },
-        {
-          "include": "#keywords"
         },
         {
           "include": "#method-call"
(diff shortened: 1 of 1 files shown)


language-support/qute/qute.tmLanguage.json
```diff
@@ -400,15 +400,23 @@
...
+          "match": "(!)",
           "name": "keyword.operator.logical.java"
         },
         {
           "match": "(\\||&)",
           "name": "keyword.operator.bitwise.java"
         },
         {
-          "match": "\\b(is|as|eq|ne|gt|ge|lt|le)\\b",
+          "match": "(?<=({#if\\s(\\w)*\\s))(gt|>|ge|>=|lt|<|le|<=)",
```
angelozerr: If you want to split match per category (logical, etc), you should change the name, otherwise use the same patch.

But I think  we should have `eq, is, ne, and, or` in `keyword.control.qute` and the other(`!= , ||, etc)`  should be `keyword.operator.qute` to have different color between them.

AlexXuChen: I split by precedence from the Qute reference guide: https://quarkus.io/guides/qute-reference#if_section, but we can do by function as well

angelozerr: As i zxplained we should not have the same color for or and ||. It means that or must be defined a given pattern and || in an another pattern

language-support/qute/qute.tmLanguage.json
```diff
@@ -380,7 +380,7 @@
           "name": "keyword.operator.assignment.bitwise.java"
         },
         {
-          "match": "(===?|!=|<=|>=|<>|<|>)",
+          "match": "(===?)",
```
angelozerr: remove this match

language-support/qute/qute.tmLanguage.json
```diff
@@ -361,7 +393,7 @@
         }
       ]
     },
-    "keywords": {
+    "keywordsToRemove": {
```
angelozerr: remove this keywordsToRemove rule.

language-support/qute/qute.tmLanguage.json
```diff
@@ -400,16 +428,20 @@
...
           "name": "keyword.operator.bitwise.java"
         },
         {
-          "match": "\\b(is|as|eq|ne|gt|ge|lt|le)\\b",
-          "name": "keyword.control.operator"
+          "match": "(?<=({#if\\s(\\w)*\\s))(gt|ge|lt|le|eq|is|ne|and|or)",
+          "name": "keyword.control.qute"
+        },
+        {
+          "match": "(?<=({#if\\s(\\w)*\\s))(>|>=|<|<=|==|!=|&&|\\|\\|)",
```
angelozerr: Move (>|>=|<|<=|==|!=|&&|\\|\\|) in section_start_if_tag

language-support/qute/qute.tmLanguage.json
```diff
@@ -400,16 +428,20 @@
...
+          "name": "keyword.operator.qute"
         },
         {
           "match": "(?<=({#for\\s(\\w)*\\s))in",
```
angelozerr: Create section_start_for_tag and defines this rule (like if operators)

language-support/qute/qute.tmLanguage.json
```diff
@@ -133,6 +139,70 @@
...
+        "2": {
+          "name": "support.constant.handlebars"
+        }
+      },
+      "patterns": [
+        {
+          "match": "(gt|ge|lt|le|eq|is|ne|and|or)",
+          "name": "keyword.control.qute"
+        },
+        {
```
angelozerr, line 169: Please defines in a new match the other operator defined in https://quarkus.io/guides/qute-reference#if_section (like  `!=`)

language-support/qute/qute.tmLanguage.json
angelozerr: @AlexXuChen please remove 

 * the full rule keywordsToRemove and defines missing operator in the `"expression"` rule defined in https://quarkus.io/guides/qute-reference#built-in-resolvers (in 2 matches one for `or` in purple and one for `?:` in black)
 * the `expression_content` rule which is not used.

language-support/qute/qute.tmLanguage.json
```diff
@@ -133,6 +139,74 @@
...
+          "name": "keyword.control"
+        },
+        "2": {
+          "name": "support.constant.handlebars"
+        }
+      },
+      "patterns": [
+        {
+          "match": "(gt|ge|lt|le|eq|is|ne|and|or)",
+          "name": "keyword.operator.qute"
```
angelozerr: It should be `keyword.control.qute` to have purple color.

language-support/qute/qute.tmLanguage.json
```diff
@@ -133,6 +139,74 @@
...
+        },
+        {
+          "match": "(!|>|>=|<|<=|==|!=|&&|\\|\\|)",
+          "name": "keyword.operator.qute"
```
angelozerr, line 171: Great, keep like this.

language-support/qute/qute.tmLanguage.json
```diff
@@ -86,7 +92,7 @@
 			"name": "string.unquoted.cdata.qute"
 		},
     "section_start_default_tag": {
-      "begin": "({)(#)((each|else|else\\sif|eval|for|if|include|insert|set|let|with|switch|case|is|when)\\b)",
+      "begin": "({)(#)((each|else|else\\sif|eval|for|include|insert|set|let|with|switch|case|is|when)\\b)",
```
angelozerr: remove `for `otherwise your rule with for will not work.

language-support/qute/qute.tmLanguage.json
```diff
@@ -198,15 +272,19 @@
         }
       },
       "patterns": [
+        {
```
angelozerr, line 275: Remove thoses matches, user tag doesn't allow those operator, it is just for #if

AlexXuChen: This isn't under the user tag match, this is under expression

angelozerr: My bad you are right. Sorry.

But matches misses some operators and keywords.

language-support/qute/qute.tmLanguage.json
```diff
@@ -198,15 +272,19 @@
...
+          "match": "or",
+          "name": "keyword.operator.qute"
```
angelozerr: Replace with keyword.control.qute

language-support/qute/qute.tmLanguage.json
```diff
@@ -198,15 +272,19 @@
         }
       },
       "patterns": [
+        {
+          "match": "or",
```
angelozerr: It misses orEmpty

language-support/qute/qute.tmLanguage.json
```diff
@@ -198,15 +272,19 @@
         }
       },
       "patterns": [
+        {
+          "match": "(or|orEmpty)",
+          "name": "keyword.control.qute"
+        },
+        {
+          "match": "(&&|\\|\\||\\?:|\\?|:)",
```
angelozerr: Is | is allowed?

AlexXuChen: The | is not a | to match, it comes after the escape of ||, which is `\\|\\|`

AlexXuChen: If I were to match | in the regex, it would be something like `\\|`, so the full match would be `(&&|\\||\\|\\||\\?:|\\?|:)`, where the | comes after &&

angelozerr:
If you write

```
{#if a orXXX b}
  
{/if}
```

or is higlighted as a keyword although it  should not (orXXX should have the same color).

image

angelozerr:
The `in` in #for is not colorized as keyword:

image

language-support/qute/qute.tmLanguage.json
angelozerr: Ok thanks for your explanation.

AlexXuChen:
@angelozerr I added word boundaries, so should work properly now:
image

language-support/qute/qute.tmLanguage.json
```diff
@@ -198,15 +272,19 @@
         }
       },
       "patterns": [
+        {
+          "match": "(or|orEmpty)",
+          "name": "keyword.control.qute"
```
angelozerr, line 277: I think \\b is missing

language-support/qute/qute.tmLanguage.json
angelozerr: I think \\b is missing

angelozerr:
It works like a charm, great job @AlexXuChen !

PR #496, opened 2022-04-29
Provide qute.native.enabled setting

angelozerr:
Provide qute.native.enabled setting

See https://github.com/redhat-developer/quarkus-ls/issues/629

### DIFF
--- a/docs/qute/README.md
+++ b/docs/qute/README.md
@@ -10,10 +10,11 @@
  ## Settings
 
  * `qute.trace.server`: Trace the communication between VS Code and the Qute language server in the Output view. Default is `off`.
+ * `qute.native.enabled`: Enable/disable Qute native image mode. Default is `false`.
  * `qute.codeLens.enabled`: Enable/disable Qute CodeLens. Default is `true`.
  * `qute.inlayHint.enabled`: Enable/disable Inlay Hint. Default is `true`.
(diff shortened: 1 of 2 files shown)


rgrunber (commented):
Just to confirm. This setting doesn't change how the quarkus build occurs for a user's project (tasks/launch files aren't touched). All it does is give the user a box to check that allows them to get better validation/assist/completions from our language server. Mainly around issues of reflection that might affect the project in the runtime.

Would it be worth adding a 1-time notification when the user first launches the binary command (`quarkusTools.buildBinary`) to inform them of this option ? Should the binary command automatically set this ?

The only change I would probably make is instead of calling it "native images", I would call it "native image mode"

angelozerr:
> Just to confirm. This setting doesn't change how the quarkus build occurs for a user's project (tasks/launch files aren't touched). All it does is give the user a box to check that allows them to get better validation/assist/completions from our language server. Mainly around issues of reflection that might affect the project in the runtime.

Exactly!

> Would it be worth adding a 1-time notification when the user first launches the binary command (quarkusTools.buildBinary) to inform them of this option ? Should the binary command automatically set this ?

The only change I would probably make is instead of calling it "native images", I would call it "native image mode"

Indeed it should be nice, we could do that in a separate PR.

angelozerr:
> The only change I would probably make is instead of calling it "native images", I would call it "native image mode"

updated.

PR #506, opened 2022-05-19
Move to vscode-languageclient 8.0.2 for Qute LS

### LINKED ISSUE #505: Move to vscode-languageclient 8.0.1 for Qute LS
Move to vscode-languageclient 8.0.1 for Qute LS. We will benefit with:

 * inlay hint support.
 * we should remove shouldLanguageServerExitOnShutdown

angelozerr:
Move to vscode-languageclient 8.0.2 for Qute LS

Fixes #505


angelozerr:
This PR requires https://github.com/redhat-developer/quarkus-ls/pull/645

rgrunber:
We should also move vscode-microprofile to vscode-languageclient 8.0.1. Currently it seems to sit at [`7.0.0`](https://github.com/redhat-developer/vscode-microprofile/blob/master/package-lock.json#L6133)

angelozerr:
> We should also move vscode-microprofile to vscode-languageclient 8.0.1. Currently it seems to sit at [7.0.0](https://github.com/redhat-developer/vscode-microprofile/blob/master/package-lock.json#L6133)

yes sure, but this PR is for Qute support. vscode-microprofile is for Quarkus support (application.properties, java support).

We can have different version of vscode-languageclient per extension). I agree with you we need to update vscode-microprofile but it is not a blocker issue for this PR.

angelozerr:
The feature which it is important to test it in this PR  is the inlay hint support.

rgrunber:
We should hold off on this until we get some clarification on https://github.com/redhat-developer/vscode-java/pull/2377#issuecomment-1138616284 . Seems to me like it would not work in Theia.

src/qute/languageServer/client.ts
```diff
@@ -2,18 +2,17 @@ import * as requirements from './requirements';
...
 import { QuteSettings } from './settings';
 import { JavaExtensionAPI } from '../../extension';
-import { QuteInlayHintsProvider } from './inlayHintsProvider';
 
-export function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI) {
+export async function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI) {
   registerVSCodeQuteCommands(context);
 
-  return requirements.resolveRequirements(api).then(requirements => {
+  return requirements.resolveRequirements(api).then(async requirements => {
```
datho7561: Since this function is now `async`, you can do:

```typescript
import { resolveRequirements } from './requirements';
// ...
export async function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI) {
  registerVSCodeQuteCommands(context);
  const requirements = await resolveRequirements(api);
  const clientOptions: LanguageClientOptions = {
  // ...
```

src/qute/languageServer/client.ts
```diff
@@ -71,59 +70,50 @@ export function connectToQuteLS(context: ExtensionContext, api: JavaExtensionAPI
...
-      );
-      // Display the experimental Qute validation pop-up if it hasn't been displayed and a Qute file is open
-      if (!hasShownQuteValidationPopUp(context)) {
-        for (const textDocument of workspace.textDocuments) {
-          if (textDocument.languageId.includes('qute')) {
-            showQuteValidationPopUp(context);
-          }
+    await quteLanguageClient.start();
+
+    bindQuteRequest('qute/template/project');
```
datho7561: These requests have been removed from the list, and should be added back
- `'qute/template/javadoc'`
- `'qute/template/generateMissingJavaMember'`

datho7561 (changes requested):

src/qute/languageServer/client.ts
angelozerr: Good catch! Fixed.

src/qute/languageServer/client.ts
angelozerr: Ouch my bad -( Fixed.

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #510, opened 2022-06-10
Fix some dependency issues reported by npm audit.

rgrunber:
- Bump vscode-extension-tester to fix underlying dependency issues with
  minimist and json-schema

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -364,9 +364,9 @@
       }
     },
     "@sindresorhus/is": {
-      "version": "4.2.0",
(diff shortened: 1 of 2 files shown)


rgrunber:
```
/home/runner/work/vscode-quarkus/vscode-quarkus/node_modules/vsce/out/package.js:136
    return (translations ?? [])
                          ^
SyntaxError: Unexpected token '?'
```
I suspect we need to start using Node 14 (instead of 12) in the builds.

PR #511, opened 2022-06-20
Add suport to auto-close curly brace before '<'

### LINKED ISSUE #502: Auto-closing of curly brace fails under certain circumstances
bracket-matching-test

I think it would be common to have something like `<element>|</element>` and want to input a `{` at the cursor (`|`) position. However, this doesn't work, and one must manually close the cursor. This works on an empty line or any line where there is whitespace to the right of the cursor. It even works in some pretty odd cases.

See https://code.visualstudio.com/api/language-extensions/language-configuration-guide#autoclosing-before for an explanation and how to override the behaviour.

JessicaJHee:
auto-close

Fixes #502 

### DIFF
--- a/language-support/qute/language-configuration.json
+++ b/language-support/qute/language-configuration.json
@@ -13,9 +13,11 @@
 		{ "open": "[", "close": "]"},
 		{ "open": "(", "close": ")" },
 		{ "open": "'", "close": "'" },
+		{ "open": "<", "close": ">" },
 		{ "open": "\"", "close": "\"" },
 		{ "open": "{!", "close": "!", "notIn": [ "comment", "string" ]}
 	],
+	"autoCloseBefore": ";:.,=}])<> \n\t",
 	"surroundingPairs": [
 		{ "open": "'", "close": "'" },
 		{ "open": "\"", "close": "\"" },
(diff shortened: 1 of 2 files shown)


language-support/qute/language-configuration.json
```diff
@@ -16,6 +16,7 @@
 		{ "open": "\"", "close": "\"" },
 		{ "open": "{!", "close": "!", "notIn": [ "comment", "string" ]}
 	],
+	"autoCloseBefore": ";:.,=}])<> \n\t",
```
rgrunber, line 20: LGTM. Just a quick question for @angelozerr : I noticed in the `autoClosingPairs` (above) we don't have the case for `<` `>` but they are in `surroundingPairs`. So selected text will be surrounded by `<` `>` but on their own it won't complete. Is that intentional, or can we fix that also.

angelozerr: To be honnest with you, I did that quickly. Don't hesitate to improve it.

rgrunber: Ok. @JessicaJHee , the `autoCloseBefore` looks fine to me. It would also apply to the other elements of `autoClosingPairs` so it's worth keeping in mind. However I couldn't find any problems. Could you please also add the case of opening `<` and closing `>` to `autoClosingPairs` ?

JessicaJHee: Yes just added!

PR #527, opened 2022-07-22
Provide HTML support for Qute HTML templates

### LINKED ISSUE #483: Provide HTML support for Qute HTML templates
Currently, Qute html templates only get Qute support. HTML validation and completion are missing. Would be nice if we could hook up the HTML LS to qute-html templates.

angelozerr:
Provide HTML support for Qute HTML templates

Fixes #483

### DIFF
--- a/language-support/qute/language-configuration.json
+++ b/language-support/qute/language-configuration.json
@@ -13,7 +13,6 @@
 		{ "open": "[", "close": "]"},
 		{ "open": "(", "close": ")" },
 		{ "open": "'", "close": "'" },
-		{ "open": "<", "close": ">" },
 		{ "open": "\"", "close": "\"" },
 		{ "open": "{!", "close": "!", "notIn": [ "comment", "string" ]}
 	],
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
   "extensionDependencies": [
     "redhat.vscode-microprofile",
     "redhat.java",
-    "vscjava.vscode-java-debug"
+    "vscjava.vscode-java-debug",
+    "vscode.html-language-features"
   ],
   "contributes": {
     "javaExtensions": [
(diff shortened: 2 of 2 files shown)


angelozerr:
To test the PR you need to instal `VSCode Insiders` which provides my contribution to vscode https://github.com/microsoft/vscode/pull/146731

Open a Qute HTML template  and you should see the HTML (completion for HTML elements, autoclose, etc) and Qute support both. Here a little demo:

QuteHTMLSupport

rgrunber (commented):
Change works very well :)

So we should ideally release vscode-quarkus 1.12.0 only once VS Code 1.70.0 (currently insiders) is released ? Do we want to set the vscode/engine to 1.70.0 for this feature ?

The only downside I'm seeing is those using an older VS Code with a newer Quarkus extension will lose even `<` auto-completion (it will be gone, but the new API from html feature also won't be present pre-1.70.0).

angelozerr:
> Change works very well :)

Great!

> So we should ideally release vscode-quarkus 1.12.0 only once VS Code 1.70.0 (currently insiders) is released ? 

I don't thing, since it is just a dclaration in package.json, it will break nothing.

> Do we want to set the vscode/engine to 1.70.0 for this feature ?

I don't know ?

>The only downside I'm seeing is those using an older VS Code with a newer Quarkus extension will lose even < auto-completion (it will be gone, but the new API from html feature also won't be present pre-1.70.0).

I think it is very annoying to have this feature because if you want to write Qute #if expression with `<`, it will generate `>` which is very annoying. I think it is better to remove that for any version.

rgrunber (approved):

angelozerr:
@rgrunber do you think we can merge the PR, or do you want to wait for 1.70 release of vscode? Do you want to update the vscode engine?

PR #536, opened 2022-08-29
Inherit microprofile-config TextMate language from vscode-microprofile

datho7561:


datho7561:
Requires https://github.com/redhat-developer/vscode-microprofile/pull/109

Apparently, the grammar injections aren't inherited when you inherit from a grammar, so the property expression grammar (`${property:DEFAULT}`) must be redefined in Quarkus as well.

rgrunber (approved):

PR #539, opened 2022-09-01
Add support for pre-releases.

rgrunber:
- Add prepare_pre_release gulp task
- Make better use of existing Jenkins parameters for FORK,
  publishPreRelease, publishToMarketPlace & publishToOVSX
- Remove unnecessary lsp4mp cloning logic

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,13 +3,6 @@
 node('rhel8'){
     stage('Checkout repos') {
         deleteDir()
-        def hasLsp4mpDir = fileExists 'lsp4mp'
-        if (!hasLsp4mpDir){
-            sh 'mkdir lsp4mp'
-        }
-        dir ('lsp4mp') {
-            git url: 'https://github.com/eclipse/lsp4mp.git'
-        }
         def hasServerDir = fileExists 'quarkus-ls'
         if (!hasServerDir){
             sh 'mkdir quarkus-ls'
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -54,10 +47,20 @@ node('rhel8'){
         }
     }
 
+    env.publishPreReleaseFlag = ""
+    if(publishPreRelease.equals('true')){
```
datho7561: In a few other cases where you are checking if a variable is equal to `'true'`, you do `"true".equals(variable)`. Maybe it's better to do that here as well?

Jenkinsfile
```diff
@@ -73,32 +76,41 @@ node('rhel8'){
...
 
-            // open-vsx Marketplace
-            sh 'npm install -g ovsx'
-            withCredentials([[$class: 'StringBinding', credentialsId: 'open-vsx-access-token', variable: 'OVSX_TOKEN']]) {
-              sh 'ovsx publish -p ${OVSX_TOKEN}' + " ${vsix[0].path}"
+            if ('true'.equals(publishToOVSX)) {
+                // open-vsx Marketplace
+                sh 'npm install -g ovsx'
+                withCredentials([[$class: 'StringBinding', credentialsId: 'open-vsx-access-token', variable: 'OVSX_TOKEN']]) {
+                sh 'ovsx publish -p ${OVSX_TOKEN}' + " ${vsix[0].path}"
```
datho7561: It'd be nice to have this line indented one more

datho7561 (changes requested):
It seems to me that it should work. Thanks, Roland! I found two things to nitpick about.

Jenkinsfile
rgrunber: Right. May as well reverse it. We don't in vscode-xml, but I think I just saw it used somewhere in the build files and decided to stick to that.

datho7561 (approved):

PR #540, opened 2022-09-06
Activation trigger should be more selective.

### LINKED ISSUE #497: Activation trigger should be more selective
The Quarkus extension and language server are triggered for [`onLanguage:java`](https://github.com/redhat-developer/vscode-quarkus/blob/master/package.json#L46). This seems excessive. I can have an invisible project with just folders and Java source files (no Maven, no Gradle), and it still starts.

quarkus-starts-on-java-project

~~At the very least we could bring it down to just :~~
```
"workspaceContains:pom.xml",
"workspaceContains:build.gradle",
"workspaceContains:settings.gradle"
```

As for a better approach, maybe vscode-java could define some command that reads the build files for presence of Quarkus metadata (eg. quarkus-bom, quarkus-maven-plugin, io.quarkus.extension, etc), and emit some command for which vscode-quarkus simply listens with https://code.visualstudio.com/api/references/activation-events#onCommand

rgrunber:
- Fixes #497
- Only start language server(s) and contribute extension functionality
  if a Quarkus project is detected

There is one annoying flaw here. When `activate()` is called, we don't know which event caused the activation, which would be nice to know.

We want to know if the activation occurred due to `onLanguage:java` because there are other activation criteria (eg. `onCommand:quarkusTools.createProject`, `workspaceContains:**/src/main/resources/application.properties`) that should definitely activate the project no matter what. So we check the known text documents, and if one has a `java` language ID, then there's a good chance, that's what activated it.

This mostly works except for the case of opening a Java file where the workspace contains an `application.properties` in the appropriate location. We will attempt to re-validate that the project is really a Quarkus project through `isQuarkusProject` when arguably, we should just activate without checking.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,8 +15,8 @@
  */
 import * as path from 'path';
 import { commands, Disposable, ExtensionContext, extensions, Terminal, TextDocument, window, workspace } from 'vscode';
-import { registerVSCodeCommands } from './commands/registerCommands';
-import { QuarkusPropertiesLanguageMismatch, QuarkusConfig } from './QuarkusConfig';
+import { registerVSCodeClientCommands, registerVSCodeCommands } from './commands/registerCommands';
+import { QuarkusConfig, QuarkusPropertiesLanguageMismatch } from './QuarkusConfig';
 import { QuarkusContext } from './QuarkusContext';
 import quarkusProjectListener from './QuarkusProjectListener';
 import { connectToQuteLS } from './qute/languageServer/client';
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -15,26 +15,42 @@
...
 export async function activate(context: ExtensionContext) {
-
   await initTelemetryService(context);
 
   QuarkusContext.setContext(context);
+
+  registerVSCodeClientCommands(context);
+  // Check if activation occured due to a 'java' language document
+  const onLanguageJavaWasActivated = workspace.textDocuments.some(doc => doc.languageId === 'java');
+  if (onLanguageJavaWasActivated) {
```
datho7561, line 43: One issue might arise if you have a rootless project or a single Java file, then you generate a Quarkus project into the file. When you go to try and use any of the other Quarkus features once the project is generated, like debugging, then it won't work.

However, this is more of an edge case, and I think it's worth it to do this the way you've coded it.

datho7561 (approved):
This seems to work pretty well to avoid launching all of Quarkus when you have individual java files or a non-quarkus project. Thanks Roland! I think we should also look into implementing this in vscode-microprofile.

PR #543, opened 2022-09-08
Add 'npm watch' task matcher, use eslint in webpack

### LINKED ISSUE #542: Add task matcher to 'npm: watch'
When launching the extension from within VS Code, a warning appears about no problem matchers being defined for 'npm: watch'. This is bad, since this means VS Code will try to launch the extension, even when there are TypeScript compile errors or eslint errors. It also means that VS Code won't wait until compilation is finished before launching.

datho7561:
Closes #542


rgrunber (approved):
Change works well for me. I can confirm with the installed recommended extensions, the launch happens once the compilation is complete.

PR #544, opened 2022-09-08
Syntax coloration for operator in #case section

### LINKED ISSUE #537: Syntax  coloration for operator of #case
Syntax  coloration for operator of #case

JessicaJHee:
image

Fixes #537 

### DIFF
--- a/language-support/qute/qute.tmLanguage.json
+++ b/language-support/qute/qute.tmLanguage.json
@@ -27,6 +27,9 @@
           "include": "#section_start_for_tag"
         },
         {
+          "include": "#section_start_case_is_tag"
+        },
+        {
           "include": "#section_start_default_tag"
         },
         {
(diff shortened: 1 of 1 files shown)


JessicaJHee:
Now works with case of {#case in} and {#case in in}:

Peek 2022-09-13 09-51

language-support/qute/qute.tmLanguage.json
```diff
@@ -245,6 +248,35 @@
...
+        }
+      },
+      "endCaptures": {
+        "1": {
+          "name": "support.constant.qute"
+        }
+      },
+      "patterns": [
+        {
+          "match": "(?<={(#case|#is)\\s)(gt|ge|lt|le|not|ne|in|ni|!in|>|>=|<|<=|!=)(?=\\s\\w+)",
```
datho7561: It might be the case that the regex engine doesn't support using `|` in a look behind. I get an error about this when I tested the regex in https://regex101.com/.

You can try something like this maybe? `(?:(?<={(#case)\s)|(?<={(#is)\s))` I got this from [this StackOverflow suggestion](https://stackoverflow.com/questions/25563891/variable-length-look-behind)

datho7561: If I do:

```
{#case  gt 10}
```
or
```
{#case gt     10}
```
then the highlighting doesn't work properly. I think it shouldn't be too hard to handle these cases.

datho7561 (changes requested):
One thing I noticed while playing with the highlighting

datho7561 (approved):
Looks good, and works well. Thanks, Jessica!

PR #550, opened 2022-09-30
Add vmargs setting for qute-ls, disable JVM logging

### LINKED ISSUE #548: Disable JVM logging by default for qute-ls
JVM logging sometimes outputs to standard out. This breaks the language support coming from the language server, because the language client and server communicate through standard in/standard out. Because of this, JVM logging should be disabled by default, and should only be enabled if the user explicitly enables it.

See https://github.com/redhat-developer/vscode-java/issues/2577

datho7561:
Closes #548

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,6 +225,12 @@
           "description": "Action performed when detected Quarkus properties have an incorrect language.",
           "scope": "window"
         },
+        "qute.server.vmargs": {
+          "type": "string",
+          "scope": "application",
+          "markdownDescription": "The vmargs to use when launching the Qute language server",
+          "default": "-Xmx100M -XX:+UseG1GC -XX:+UseStringDeduplication -Xlog:disable"
+        },
         "qute.templates.languageMismatch": {
           "type": "string",
           "enum": [
--- a/src/qute/languageServer/quteServerStarter.ts
+++ b/src/qute/languageServer/quteServerStarter.ts
@@ -30,12 +30,18 @@ function prepareParams(): string[] {
     }
   }
 
-  const vmargs = workspace.getConfiguration("quarkus.tools").get("server.vmargs", '');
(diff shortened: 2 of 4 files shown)


src/qute/languageServer/quteServerStarter.ts
```diff
@@ -30,13 +30,19 @@ function prepareParams(): string[] {
     }
   }
 
-  const vmargs = workspace.getConfiguration("quarkus.tools").get("server.vmargs", '');
```
rgrunber, line 33: Just to confirm, this basically did nothing right ? Unless a user actually defined the property manually, which we didn't really "broadcast" via package.json ..

package.json
```diff
@@ -225,6 +225,12 @@
           "description": "Action performed when detected Quarkus properties have an incorrect language.",
           "scope": "window"
         },
+        "qute.server.vmargs": {
+          "type": "string",
+          "scope": "machine",
+          "markdownDescription": "The vmargs to use when launching the Qute language server",
+          "default": "-Xmx100M -XX:+UseG1GC -XX:+UseStringDeduplication"
```
rgrunber: Any reason not to add `-Xlog:disable` here as the default ?

package.json
```diff
@@ -225,6 +225,12 @@
           "description": "Action performed when detected Quarkus properties have an incorrect language.",
           "scope": "window"
         },
+        "qute.server.vmargs": {
```
rgrunber, line 228: Also, don't forget to update https://github.com/redhat-developer/vscode-quarkus/blob/master/docs/qute/README.md#settings .

src/qute/languageServer/quteServerStarter.ts
datho7561: Yes, this did nothing.

package.json
datho7561: I'll add it in

PR #560, opened 2022-12-02
Fix check to determine if a Quarkus project is open

### LINKED ISSUE #559: "Create project" fails when there are no folders and one java file open

datho7561:
Closes #559

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -144,6 +144,9 @@ export async function getJavaExtensionAPI(): Promise<JavaExtensionAPI> {
   return Promise.resolve(api);
 }
 async function isQuarkusProject(): Promise<boolean> {
+  if (!workspace.workspaceFolders) {
+    return false;
+  }
   for (const ws of workspace.workspaceFolders) {
     const buildFileUris = await getFilePathsFromWorkspace(ws, "**/{pom.xml,build.gradle}");
     for (const uri of buildFileUris) {


rgrunber (approved):
Ok, so I finally remembred why I had `onLanguageJavaWasActivated` on the Quarkus end.

https://github.com/redhat-developer/vscode-quarkus/blob/c31275be731cc047b5eccdc005253e017c1d4c98/package.json#L36-L51

If the activation happens for any reason other than `onLanguage:java`, we should honour it. The only time we need to do more work, is if the activation was due to `onLanguage:java`. In fact, that's why the commands work.

Similarly, I think in vscode-microprofile, we should not do extra work when the activation is clearly something that requires MP. We don't have commands in vscode-microprofile so we don't run into the same problems, but I still think we should limit our check.

datho7561:
Just to confirm, you think we should have the same logic in vscode-microprofile as we have here?

rgrunber:
Yes. I think we're still fine if someone opens a project that has a microprofile config file and we confirm it really is a Java project, but if we were to create a command in vscode-microprofile that should trigger activation, then we would likely need to implement the same check.

datho7561:
What do you think about exporting an api from vscode-microprofile with a function that can trigger the language server to start? This way, we can avoid starting the language server until it's needed (eg. a `pom.xml` or `build.gradle` file is detected in the workspace). Then, if vscode-quarkus (or other extensions that depend on vscode-microprofile) need the language server, they can activate the language server.

rgrunber:
> What do you think about exporting an api from vscode-microprofile with a function that can trigger the language server to start? This way, we can avoid starting the language server until it's needed (eg. a `pom.xml` or `build.gradle` file is detected in the workspace). Then, if vscode-quarkus (or other extensions that depend on vscode-microprofile) need the language server, they can activate the language server.

It feels like the best place for such an API would be in vscode-java. It should inform whether something is a Java project.

PR #566, opened 2022-12-12
Fix test failures due to VSCode API change

datho7561:


rgrunber:
https://github.com/microsoft/vscode/pull/165783#issuecomment-1347904879 . So this workaround may just be a temporary measure.

PR #571, opened 2023-01-16
Use the vsce in the `@vscode` namespace

datho7561:
- Remove the old `vsce` from the `package-lock.json`

See https://github.com/redhat-developer/vscode-xml/pull/857


rgrunber (approved):
LGTM

PR #575, opened 2023-01-26
Migrate from find-java-home to jdk-utils

### LINKED ISSUE #569: Migrate from `find-java-home` to `jdk-utils`
This would mean that we use the same strategy as in vscode-java.

JessicaJHee:
Fixes #569


datho7561:
Make sure to replace the repository.engineering.redhat.com/nexus/repository URLS in the package-lock.json with the corresponding npm ones, so that non-Red Hat developers working on this project can download the dependencies. (you should be able to do this with a quick search and replace).

datho7561:
Other than what I mentioned, this PR looks good. It should probably never need to be run, since vscode-java ships with a JDK on most platforms, and if vscode-java doesn't detect a JDK, then it won't launch, which should prevent vscode-quarkus from launching.

src/qute/languageServer/requirements.ts
```diff
@@ -33,48 +33,49 @@ export async function resolveRequirements(api: JavaExtensionAPI): Promise<Requir
...
         }
+        return javaHome;
+    }
+    // No settings, let's try to detect as last resort.
+    const javaRuntimes = await findRuntimes({ withVersion: true, withTags: true });
+    if (javaRuntimes.length) {
+        sortJdksBySource(javaRuntimes);
+        javaHome = javaRuntimes[0].homedir;
+    } else {
+        openJDKDownload("Java runtime could not be located. Please download and install Java or use the binary server.");
```
datho7561: `throw` the result of this function, similar to how it's done earlier in the function

datho7561 (changes requested):
Sorry, caught something small

datho7561 (approved):
Sorry, forgot to get back to this. Looks good! Thanks, Jessica!

PR #577, opened 2023-01-31
Call `quteLanguageServer.stop()` when deactivating

### LINKED ISSUE #576: qute-ls isn't stopped by vscode-quarkus
qute-ls is left running after closing VS Code, instead of being stopped when VS Code is closed. It remains running until the auto shut off mechanism detects that the parent process is dead, ~30 seconds later.

datho7561:
Closes #576


angelozerr:
Thanks @datho7561

PR #578, opened 2023-02-01
Update .vscode/tasks.json to run Debug Quarkus command with missing mvnw

### LINKED ISSUE #572: when no ./mvnw exist cannot debug quarkus
if no mvnw exist in project debug quarkus fails trying to run `./mvnw quarkus:dev`

...could run `mvn quarkus:dev` to rely on path ...or at least tell user how they configure which command it will run.

JessicaJHee:
Fixes #572 

### DIFF
--- a/src/utils/tasksUtils.ts
+++ b/src/utils/tasksUtils.ts
@@ -17,6 +17,7 @@ import * as _ from 'lodash';
 
 import { tasks, ProcessExecution, ShellExecution, Task, TaskExecution, WorkspaceFolder, CustomExecution } from 'vscode';
 import { BuildSupport } from '../buildSupport/BuildSupport';
+import { getFilePathsFromFolder } from '../utils/workspaceUtils';
 
 export async function getQuarkusDevTaskNames(workspaceFolder: WorkspaceFolder, projectBuildSupport: BuildSupport) {
   const quarkusDevTaskDefinitions: Task[] = await getQuarkusDevTasks(workspaceFolder, projectBuildSupport);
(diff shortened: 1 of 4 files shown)


rgrunber (commented):
If I take your change and a quarkus project (eg. [config-quickstart](https://github.com/quarkusio/quarkus-quickstarts/tree/main/config-quickstart)) with no `.vscode/` folder and just call the `Debug Quarkus` command, it generates everything as before, but the debugging doesn't start.

It looks like the last line of `startDebugging` :

`await debug.startDebugging(workspaceFolder, debugConfig);` never runs.

I think part of the issue is that `getQuarkusDevDebugConfig` is also used by `waitUntilTaskExists` as the last step up creating the launch config in `createLaunchConfig`.

I also noticed `shouldUpdateTaskCommand()` is returning undefined even when the wrapper command is still there. I'd have a look at `getWrapperPathFromBuildFile` as I think that isn't finding the wrapper.

JessicaJHee:
@rgrunber you were correct! I needed to use another way to find the wrapper command. It should work fine now

src/utils/tasksUtils.ts
```diff
@@ -83,6 +84,26 @@ function isQuarkusDevTask(task: Task, projectBuildSupport: BuildSupport): boolea
...
+
+  const execution: ProcessExecution | ShellExecution | CustomExecution = task.execution;
+
+  if (!execution || !(execution as ShellExecution)) {
+    return false;
+  }
+
+  const shellExecution: ShellExecution = execution as ShellExecution;
+
+  if (shellExecution
```
rgrunber: Should be able to simplify this a bit with https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html .

Also I think instead of `if (array.length > 0)` you can simply use `if(array)`

src/utils/tasksUtils.ts
```diff
@@ -76,13 +77,31 @@ function isQuarkusDevTask(task: Task, projectBuildSupport: BuildSupport): boolea
...
+  const execution: ProcessExecution | ShellExecution | CustomExecution = task.execution;
+
+  if (!execution || !(execution as ShellExecution)) {
+    return false;
+  }
+
+  const shellExecution: ShellExecution = execution as ShellExecution;
+
+  if (shellExecution?.commandLine
+    && (shellExecution.commandLine.includes(projectBuildSupport.getWrapper())
```
rgrunber: If I'm on Linux and I delete the `mvnw` script, this logic will not update the task because `mvnw.cmd` exists. But on Linux `mvnw.cmd` can't be used. Try deleting just `mvnw` and it'll fail because this condition allows it. Same applies if I'm on Windows and delete the `mvnw.cmd` script.

src/wizards/debugging/TaskCreator.ts
```diff
@@ -124,7 +124,7 @@ export class TaskCreator {
    */
   private async addTask(taskDefinition: TaskDefinition): Promise<void> {
     const tasksJson = workspace.getConfiguration('tasks', this.workspaceFolder.uri);
-    const tasks: TaskDefinition[] = tasksJson.get<TaskDefinition[]>('tasks');
+    const tasks: TaskDefinition[] = tasksJson.get<TaskDefinition[]>('tasks').filter(task => task['label'] != taskDefinition['label']);
```
rgrunber, line 127: I think the same problem exists at https://github.com/redhat-developer/vscode-quarkus/blob/2f275047459fe3791874a29e92c8657e4e9f8f98/src/wizards/debugging/LaunchConfigCreator.ts#L89-L90 . This could be a separate bug since it has nothing to do with your change, but it would be good to fix it.

You could actually reproduce it by just removing the tasks.json file because that will cause the launch to be regenerated either way.

rgrunber (commented):
Overall, this works in the general case. Just some things to polish up and one thing to fix.

One thing to keep in mind that I don't think we need to deal with here is priority. For example, let's say you can't locate the wrapper in the project folder, so you use the default, which works. What happens if the wrapper is put back/re-introduced. Should we detect it and refresh again ? We can address this an another issue if we really feel we should handle this.

src/wizards/debugging/LaunchConfigCreator.ts
```diff
@@ -86,7 +86,8 @@ export class LaunchConfigCreator {
 
   private async addLaunchConfig(): Promise<void> {
     const launchJson = workspace.getConfiguration('launch', this.workspaceFolder.uri);
-    const configurations: DebugConfiguration[] = launchJson.get<DebugConfiguration[]>('configurations');
+    const configurations: DebugConfiguration[] = launchJson.get<DebugConfiguration[]>('configurations')
+      .filter(task => task['preLaunchTask'] != this.quarkusBuildSupport.getQuarkusDevTaskName(this.workspaceFolder, this.projectFolder));
```
rgrunber: Is there any reason to use `preLaunchTask` as opposed to `name` ? If someone wants to have a launch that uses our pre-defined `quarkus:dev` task, that's fine. No reason that multiple can't exist. The only thing is we shouldn't allow duplicate launches, which we can interpret to mean launches with the same `name`.

rgrunber (commented):
Nice. This even gives priority to the project wrapper (if it exists) ! Just one small thing to address and I think we can merge.

rgrunber (approved):
Looks mostly fine to me. If @datho7561 is fine with it, feel free to merge.

datho7561 (approved):
This works well for me. Thanks, Jessica!

PR #580, opened 2023-02-03
Major bug with definition of user tag

### LINKED ISSUE #579: Major bug with definition of user tag
I don't know if it is because of the new vscode language client, but there is a major bug with definition for user tag. To reproduce the problem, do Ctrl+Click on an user tag in qute template without click on it (to avoid opening the file)

image

And you will see in the trace an infinite of didOpen, didClose of the user tag file name, which report  a lot of diagnostics (because diagnostics is done on didOpen).

To stop those didOpen/didClose, you need to open the user tag file name.

angelozerr:
Major bug with definition of user tag

Fixes #579


datho7561 (approved):
Works great! Thanks, Angelo!

PR #582, opened 2023-02-03
Enable eslint's `no-inferrable-types`

### LINKED ISSUE #519: Enable "no-inferrable-types" ESLint rule
As of now, we have the following ESLint rule disabled: https://github.com/redhat-developer/vscode-quarkus/blob/master/.eslintrc.json#L26. I think that this rule is helpful and should be enabled; it tends to make code easier to read since there is less clutter from the type hints. We didn't do this change in the initial PR, since it would take a lot of time to clean up the codebase so that ESLint doesn't complain, and this would mean it would have taken longer to migrate to ESLint.

datho7561:
Fixes #519

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,7 +23,6 @@
       }
     ],
     "@typescript-eslint/no-namespace": "off",
-    "@typescript-eslint/no-inferrable-types": "off",
     "@typescript-eslint/no-explicit-any": "off",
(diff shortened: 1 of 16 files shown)


datho7561:
@JessicaJHee if you have time, do you think you could review this PR? This is more of a nice to have, so no worries if not.

JessicaJHee (approved):
Looks good to me and makes sense!

PR #584, opened 2023-02-07
Read files without triggering onDidOpenTextDocument.

rgrunber:
- vscode.workspace.openTextDocument(..) will trigger the vscode.workspace.onDidOpenTextDocument(..) callback
- Some extensions (eg. vscode-java) may relay on this (particularly for pom files) in order to activate other functionality generating a false positive (ie. pom file was "opened")

I was able to reproduce this by having vscode-quarkus + vscode-pde installed (with all associated dependencies), and then opening eclipse.jdt.ls as a project, and ensuring that some Java source file is opened. Once the project is loaded, you should see a flurry of notifications opening and disappearing very quickly. They all appear to be from the vscode-java [dependency analytics prompt](https://github.com/redhat-developer/vscode-java/blob/5c0123f7cfd88086e187112c0a616ac4e90767c6/src/recommendation/dependencyAnalytics.ts#L27).

dep-analytics-popup

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,6 +28,7 @@ import { initTelemetryService } from './utils/telemetryUtils';
 import { getFilePathsFromWorkspace } from './utils/workspaceUtils';
 import { WelcomeWebview } from './webviews/WelcomeWebview';
 import { createTerminateDebugListener } from './wizards/debugging/terminateProcess';
+import { readFile } from 'fs-extra';
 
 // alias for vscode-java's ExtensionAPI
 export type JavaExtensionAPI = any;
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -160,8 +161,8 @@ async function isQuarkusProject(): Promise<boolean> {
   for (const ws of workspace.workspaceFolders) {
     const buildFileUris = await getFilePathsFromWorkspace(ws, "**/{pom.xml,build.gradle}");
     for (const uri of buildFileUris) {
-      const doc = await workspace.openTextDocument(uri);
-      if (doc.getText().search("io.quarkus") > 0) {
+      const doc = readFileSync(uri.fsPath);
```
datho7561: It might be better to read the file asynchronously. I think you you use the function returned by `promisify(readFile)` that should work?

datho7561:
I don't think I see popups, but I do see something on the left of the bottom bar. Is this what you are mentioning? If so, then this PR does prevent that from happening.

src/extension.ts
```diff
@@ -157,14 +158,21 @@ async function isQuarkusProject(): Promise<boolean> {
   if (!workspace.workspaceFolders) {
     return false;
   }
-  for (const ws of workspace.workspaceFolders) {
+  Promise.all(workspace.workspaceFolders.map(async ws => {
```
datho7561: Make sure to return the value of this promise

datho7561 (approved):
Looks good and works well. Thanks, Roland!

PR #585, opened 2023-02-22
Surround with section/comments command

### LINKED ISSUE #581: Surround with section / comments
Liek we have for vscode-xml surround with tags https://github.com/redhat-developer/vscode-xml/blob/main/docs/Refactor.md#surround-with-tags-wrap

It should really nice to have a surround with section.

Given this Qute template:

`{class}`

It should be nice to to select this expression execute the command surround with section which will generate that:

`{#if condition}{class}{\if}`

@FroMage, @mkouba in XML the surround with tags is very nice, I think we can have similar feature for Qute and having a surroud with comments too.

JessicaJHee:
Requires https://github.com/redhat-developer/quarkus-ls/pull/801 

Fixes #581


package-lock.json
```diff
@@ -599,9 +599,9 @@
       "dev": true
     },
     "@types/vscode": {
-      "version": "1.65.0",
-      "resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.65.0.tgz",
-      "integrity": "sha512-wQhExnh2nEzpjDMSKhUvnNmz3ucpd3E+R7wJkOhBNK3No6fG3VUdmVmMOKD0A8NDZDDDiQcLNxe3oGmX5SjJ5w==",
+      "version": "1.75.1",
```
angelozerr: Why you need to do this change?

package.json
```diff
@@ -175,6 +175,18 @@
...
         "title": "Enable Qute Validation",
         "icon": "$(eye-closed)"
+      },
+      {
+        "command": "qute.refactor.surround.with.comments",
+        "title": "Qute: Surround with Comments"
+      },
+      {
+        "command": "qute.refactor.surround.with.cdata",
+        "title": "Qute: Surround with CDATA"
```
angelozerr: `Qute: Surround with Unparsed Character Data`

src/qute/commands/registerCommands.ts
```diff
@@ -1,10 +1,11 @@
 import * as path from 'path';
+import * as vscode from 'vscode';
 import { TextEncoder } from "util";
-import { commands, ConfigurationTarget, ExtensionContext, Position, Range, Selection, TextDocument, TextEditor, Uri, window, workspace, WorkspaceConfiguration } from "vscode";
-import { ConfigurationItem, Location } from "vscode-languageclient";
+import { commands, ConfigurationTarget, ExtensionContext, Position, Range, Selection, SnippetString, TextDocument, TextEditor, Uri, window, workspace, WorkspaceConfiguration, WorkspaceEdit } from "vscode";
```
angelozerr: Clean the import by removing SnippetString and WorkspaceEdit

angelozerr:
@JessicaJHee at the end of https://github.com/redhat-developer/vscode-quarkus/blob/master/docs/qute/TemplateSupport.md it would be nice that you add a Surround command for 3 types (comments, section, unparsed character data) by explaining how to use the command and add a little demo. For the demo please use the white theme to be consistent with other demos.

docs/qute/TemplateSupport.md
```diff
@@ -111,3 +111,27 @@ Qute support displays the Java type of the parameter names as an inlay hint.
...
+
+This command provides the ability to select Qute content and surround it some content. To execute this command you can use the command palette (`Ctrl+Shift+P`) and type `Surround`.
+
+If you prefer using keyboard to process the command, you need to associate this command with a keybinding. See [Keyboard Shortcuts editor](https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-shortcuts-editor) for more informations.
+
+### Surround with Section
+
+Surrounds the selected content with a section and completion will be triggered. You can also do an empty section on the section you would like to surround and the correct section will selected.
+
+![Qute Surround with Section](images/QuteTemplateSurroundWithSection.gif)
```
angelozerr: You don't apply #each in the demo, it is because you are waiting for my PR?

package.json
```diff
@@ -175,6 +175,18 @@
         "command": "qute.validation.enabled.toggle.on",
         "title": "Enable Qute Validation",
         "icon": "$(eye-closed)"
+      },
+      {
+        "command": "qute.refactor.surround.with.comments",
```
angelozerr, line 180: You need to write some filter because today:

 * your command are availablefor any files (ex : xml file)
 * your command is available even if the Qute LS is not started.

I suggest that you see XML https://github.com/redhat-developer/vscode-xml/blob/7accd8a0bf5d15aa0955a281b6edc9caacd2e0c4/package.json#L815

```
{
          "command": "xml.refactor.surround.with.comments",
          "when": "editorLangId in xml.supportedLanguageIds && XMLLSReady"}
}
```

For XMLLSReady, see https://github.com/redhat-developer/vscode-xml/blob/main/src/client/xmlClient.ts#L43

For xml.supportedLanguageIds, see https://github.com/redhat-developer/vscode-xml/blob/a17945188ade7f39a3d8ee0fbc75443db5ac3200/src/extension.ts#L46

docs/qute/TemplateSupport.md
JessicaJHee: I will update the demo :+1:

package.json
datho7561, line 180: You can already surround with comment using `Ctrl+Shift+/`. Is it worthwhile having a separate command to do this/does it do something different in certain cases?

angelozerr: The Qute comment uses the AST, so for instance if you wish comment full section, you set the cursor to the section and apply comment which will comment the full section content.

JessicaJHee: I have made some changes for this issue but I'm not sure why the command still exists even under those conditions?

docs/qute/TemplateSupport.md
angelozerr: I think for the surround demo section, it should be better to show a concrete helpfull usecase. If you wish you can use my 2 following demos:

 * with #for
SurroundWithForSection

 * with #if

SurroundWithIfSection

package.json
angelozerr: Your when is declared in the wrong place. You need to redeclare the command in command palette section like https://github.com/redhat-developer/vscode-quarkus/blob/cc2e27697dca8c9cfbfff5b9c7bea885f5151fa6/package.json#L340

package.json
```diff
@@ -175,6 +175,21 @@
...
+        "title": "Qute: Surround with Comments"
+      },
+      {
+        "command": "qute.refactor.surround.with.cdata",
+        "when": "quarkusProjectExistsOrLightWeight && editorLangIdSupportsQute && QuteLSReady",
+        "title": "Qute: Surround with Unparsed Character Data"
+      },
+      {
+        "command": "qute.refactor.surround.with.section",
+        "when": "quarkusProjectExistsOrLightWeight && editorLangIdSupportsQute && QuteLSReady",
```
angelozerr: where you defineeditorLangIdSupportsQute? I suggest that you do like vscode-xml by writing `editorLangId in qute.supportedLanguageIds` and you fill  `qute.supportedLanguageIds` like https://github.com/redhat-developer/vscode-xml/blob/a17945188ade7f39a3d8ee0fbc75443db5ac3200/src/extension.ts#L46

JessicaJHee: It's defined here: https://github.com/redhat-developer/vscode-quarkus/blob/2f275047459fe3791874a29e92c8657e4e9f8f98/src/qute/commands/registerCommands.ts#L210
and it's used to display the validation icon for Qute templates

package.json
datho7561: Okay. That makes sense, and sound helpful.

src/qute/commands/registerCommands.ts
```diff
@@ -31,6 +31,7 @@ export function registerVSCodeQuteCommands(context: ExtensionContext) {
 }
 
 export function registerQuteExecuteWorkspaceCommand(context: ExtensionContext, languageClient: LanguageClient) {
```
datho7561: Since this registers more commands than just `QuteClientCommandConstants.EXECUTE_WORKSPACE_COMMAND` now, I think you should rename it to something like `registerQuteLSDependentCommands`

datho7561 (approved):
Works very well! One small comment about naming, but feel free to merge after addressing it.

PR #589, opened 2023-02-27
Wait on jdt.ls standard mode before issuing delegate commands

### LINKED ISSUE #586: request failed on every java file load
I get a notification saying requests failed.

log has this in Qute support view:

```
[Error - 8:17:32 PM] Request textDocument/inlayHint failed.
[object Object]
[Error - 8:17:33 PM] Request textDocument/inlayHint failed.
[object Object]
[Error - 8:17:34 PM] Request textDocument/inlayHint failed.
[object Object]
[Error - 8:17:34 PM] Request textDocument/inlayHint failed.
[object Object]
[Error - 8:17:34 PM] Request textDocument/inlayHint failed.
[object Object]
```

datho7561:
Use the API to wait until jdt.ls is running before issuing commands to it. This bug is happening, since jdt.ls is no longer guaranteed to be running when vscode-quarkus launches.

Fixes #586

### DIFF
--- a/src/qute/languageServer/client.ts
+++ b/src/qute/languageServer/client.ts
@@ -57,8 +57,10 @@ export async function connectToQuteLS(context: ExtensionContext, api: JavaExtens
   };
 
   function bindQuteRequest(request: string) {
-    quteLanguageClient.onRequest(request, async (params: any) =>
-      <any>await commands.executeCommand("java.execute.workspaceCommand", request, params)
+    quteLanguageClient.onRequest(request, async (params: any) => {
+      await api.serverReady();
+      return commands.executeCommand("java.execute.workspaceCommand", request, params);
+    }
     );
   }
 


datho7561:
I'll make a similar change to vscode-microprofile to prevent the popup there.

rgrunber:
Thanks for catching these. Feel free to merge when they're ready. Do the inlay hints show up sooner if we respond (as opposed to just ignoring the request) ?

datho7561:
I'm not sure I completely understand what you are saying.

I looked into sending a blank response to qute-ls if it requests info from jdt.ls when jdt.ls isn't started yet, but that also resulted in errors. I think this is what you mean?

rgrunber:
Yeah, I was curious if just ignoring the request, since the server isn't ready, was any different than waiting for the server and potentially responding late.

datho7561:
No, when I did a quick test, the code lens appear pretty quickly after jdt.ls starts in either case.

PR #592, opened 2023-03-07
Remove vscode-extension-tester-native

### LINKED ISSUE #574: `vscode-extension-tester-native` is deprecated
See https://www.npmjs.com/package/vscode-extension-tester-native

We should assess if we need it. If we remove it, I think we can avoid the dependency on opencv for the tests.

datho7561:
This prevents opencv from being compiled during the UI tests.

Closes #574

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -181,73 +181,6 @@
         "fastq": "^1.6.0"
       }
     },
-    "@nut-tree/libnut": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/@nut-tree/libnut/-/libnut-2.1.0.tgz",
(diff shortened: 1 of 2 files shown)


datho7561:
@JessicaJHee if you have time, do you think you could review this PR? This one is more important than https://github.com/redhat-developer/vscode-quarkus/pull/582

JessicaJHee:
If I understand correctly, we can remove this because we don't use native dialogs? And I'm also wondering if the test failures on this PR are relevant?

datho7561:
Let me fix the test failures, those are my bad

datho7561:
> we don't use native dialogs?

I think so. According to https://www.npmjs.com/package/vscode-extension-tester-native it's deprecated and there is a workaround if we need the functionality it provides.

datho7561:
Tests should work now (except for https://github.com/redhat-developer/vscode-quarkus/pull/603)

datho7561:
@rgrunber if you have time, could you please review this PR? If not, I understand, this PR isn't that important

PR #597, opened 2023-03-17
Show validation errors for Qute project without opening a Qute template file

angelozerr:
Show validation errors for Qute project without opening a Qute template file

See https://github.com/redhat-developer/quarkus-ls/issues/813


datho7561 (approved):
Looks good to me!

PR #603, opened 2023-04-04
Fix extension duplicates testcase.

rgrunber:
- Move down the quick-pick list one less than the maximum visible items
- Also update vscode-extension-tester 4.2.5 -> 5.5.2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -427,15 +427,15 @@
       "dev": true
     },
     "@types/cacheable-request": {
-      "version": "6.0.2",
-      "resolved": "https://registry.npmjs.org/@types/cacheable-request/-/cacheable-request-6.0.2.tgz",
-      "integrity": "sha512-B3xVo+dlKM6nnKTcmm5ZtY/OL8bOAOd2Olee9M1zft65ox50OzjEHW91sDiU9j6cvW8Ejg1/Qkf4xd2kugApUA==",
+      "version": "6.0.3",
+      "resolved": "https://registry.npmjs.org/@types/cacheable-request/-/cacheable-request-6.0.3.tgz",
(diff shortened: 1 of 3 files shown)


src/test/vscodeUiTest/suite/projectGenerationTest.ts
```diff
@@ -528,7 +528,7 @@ describe('Project generation tests', function () {
     await wizard.next();
 
     const allQuickPickInfo: QuickPickItemInfo[] = await wizard.getAllQuickPickInfo();
-    const allLabels: string[] = allQuickPickInfo.map((info) => info.label);
+    const allLabels: string[] = allQuickPickInfo.filter(info => info.label != "").map((info) => info.label);
```
datho7561: If you make this change in `requestUtils.ts#removeDuplicateArtifactIds` instead of here, does that also fix the tests?

rgrunber: I could give it a try. I still don't understand why this is required for GH actions to pass, and not locally, but I was clearly seeing empty string entries in the list.

rgrunber:
@datho7561 so the "empty" label entries that needed filtering come from the actions that moves the selection down one entry in the quick pick menu. Supposing the quick-pick can display N entries, it first calls the key-down action N-1 times (the 1st entry is already selected), and then all subsequent moves are done N times to ensure the we're always reading in new entries. For some reason the first entry just comes back as empty.

I noticed that when building the list of labels, we check to only added new ones, so there shouldn't be any harm in moving the key-down some non-optimal amount (eg. 1,2, .. N-1) . Not sure why it's an issue in the test machine and not locally.

The vscode-extension-tester update isn't needed, and doesn't seem related but we might as well do it, unless you prefer to add it to your vscode-extension-tester-native removal.

datho7561:
I'm okay with updating the extension tester in this PR.

datho7561 (approved):
Looks good. Thanks for investigating this, Roland!

PR #606, opened 2023-04-18
Check out the tag of quarkus-ls specified by `QUARKUS_LS_TAG`

### LINKED ISSUE #602: Specify the tag of `quarkus-ls` to to bundle with the extension when releasing
In vscode-microprofile, the Jenkins pipeline has a parameter to specify which tag of lsp4mp to bundle with the extension when releasing. This is helpful in the case that you don't want to use the latest commit on `quarkus-ls` in the release. For instance, I was planning on merging the upversion PR https://github.com/redhat-developer/quarkus-ls/pull/854 before triggering the vscode-quarkus build. However, if I had merged it, the vscode-quarkus build would have bundled the `quarkus-ls` with the updated version number instead of the correct version.

See https://github.com/redhat-developer/vscode-microprofile/blob/master/Jenkinsfile#L11

datho7561:
References a new parameter `QUARKUS_LS_TAG`, and checks out this tag of quarkus-ls after cloning the repo. Copied from vscode-microprofile.

Closes #602

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,9 @@ node('rhel8'){
             sh 'mkdir quarkus-ls'
         }
         dir ('quarkus-ls') {
+            echo "Checking out quarkus-ls ${params.QUARKUS_LS_TAG}"
             git url: 'https://github.com/redhat-developer/quarkus-ls.git'
+            sh "git checkout ${params.QUARKUS_LS_TAG}"
         }
         def hasClientDir = fileExists 'vscode-quarkus'
         if (!hasClientDir) {


Jenkinsfile
```diff
@@ -8,7 +8,9 @@ node('rhel8'){
             sh 'mkdir quarkus-ls'
         }
         dir ('quarkus-ls') {
+            echo "Checking out quarkus-ls ${params.QUARKUS_LS_TAG}"
             git url: 'https://github.com/redhat-developer/quarkus-ls.git'
+            sh "git checkout ${params.QUARKUS_LS_TAG}"
```
rgrunber, line 13: Could something like `git url: 'https://github.com/redhat-developer/quarkus-ls.git', branch: ${params.QUARKUS_LS_TAG}` also work ?

rgrunber: Also is it worth using ${params.FORK} instead of redhat-developer, or is that just overkill.

datho7561: I don't know if using the `branch` param will work, I just copied this from vscode-microprofile. It might not be a good idea to use `params.FORK` for cloning `quarkus-ls`, since you might want to use vscode-quarkus and quarkus-ls from different github organizations.

rgrunber (approved):
Ok, then let's just keep this similar to vscode-microprofile.

PR #607, opened 2023-04-28
Improve information collected for selected extensions.

rgrunber:
- "Create Project" wizard extension data should be sent as a list
- "Add Extensions" quick-pick should also report extension data


rgrunber:
The current "Add Extensions" command seemed to always return success so I don't think we're losing much by not using it to wrap the command. Maybe we could check the Terminal exit status, though I noticed we don't wait for the terminal command to complete before returning from it.

datho7561 (approved):
Looks good! Thanks, Roland!

PR #621, opened 2023-07-26
Handle telemetry/event coming from the Qute language server.

rgrunber:
- Requires https://github.com/redhat-developer/quarkus-ls/pull/917 (or any `telemetry/event` sent from the server)


datho7561 (approved):
Looks good. Thanks, Roland!

PR #642, opened 2024-02-01
Fix UI tests

### LINKED ISSUE #626: Node 16 is EOL
Node 16 will also be EOL (no maintenance) on September 11 2023. As a result, we should upgrade to Node 18.

A part of this migration is updating the version of npm used and updating `package-lock.json`. npm v7, which ships with Node 16, changed the format of this file. This might cause some complications with `npm-force-resolutions`, which is a library we use to for all instances of a library to a specified version in the lockfile.

datho7561:
- Update vscode-extension
- Fix command line args to the extension tester CLI
- Update to Node 18

Fixes #626

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,9 +23,9 @@ jobs:
         uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '14.x'
+          node-version: '18.15'
       - name: Install vsce
-        run: npm i -g vsce
+        run: npm i -g @vscode/vsce
       - name: Install dependencies
         run: npm i
       - name: Compile extension
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ node('rhel8'){
     }
 
     stage('Install requirements') {
-        def nodeHome = tool 'nodejs-14.19.1'
+        def nodeHome = tool 'nodejs-18.16.1'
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -23,9 +23,9 @@ jobs:
         uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '14.x'
+          node-version: '18.15'
```
fbricon, line 26: you have to update the jenkinsfile as well

fbricon: might as well use 18.16.1, to align with Jenkinsfile

fbricon: ... unless 18.15 is the version supported by vscode

fbricon:
> WARNING: You are using the outdated VS Code version '1.85.1'. The latest stable version is '1.86.0'.

So that's governed by extension tester?

fbricon:
Thanks @datho7561!

.github/workflows/tests.yml
datho7561: 18.15 is the version supported by vscode-extension-tester. It's not available (yet) on our Jenkins instance, but I'm going to add it. Luckily, our Jenkins script doesn't seem to run the UI tests (likely because we don't want the release to fail due to a flaky UI test), so using a newer version shouldn't be a problem for the time being.

PR #644, opened 2024-02-01
Use NodeJS 18.15.0 on Jenkins

### LINKED ISSUE #643: Use Node 18.15 for Jenkins builds
We use Node 18.15 to test the extension on GitHub Actions. It would be nice to also use node 18.15 to build the extension on Jenkins instead of Node 18.16.

Node 18.15.0 is now set up on our Jenkins instance, so we can use it in the script.

datho7561:
Closes #643

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ node('rhel8'){
     }
 
     stage('Install requirements') {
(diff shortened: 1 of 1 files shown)


fbricon:
Thanks @datho7561!

PR #645, opened 2024-02-01
Enable dependabot

datho7561:
This enables dependabot to create PRs that update npm dependencies to the latest version.

This has been helpful in keeping dependencies up to date in vscode-openshift-toolkit, and I think it could be helpful here as well.


fbricon:
Thanks @datho7561!

PR #657, opened 2024-02-05
Replace vulnerable pom-parser with fast-xml-parser in tests, and other updates

fbricon:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "bugs": "https://github.com/redhat-developer/vscode-quarkus/issues",
   "engines": {
-    "vscode": "^1.65.0"
+    "vscode": "^1.82.0"
   },
   "galleryBanner": {
     "color": "#d8ebff",
@@ -34,8 +34,6 @@
     "disableAutoBuild": 1
   },
   "activationEvents": [
-    "onCommand:quarkusTools.createProject",
-    "onCommand:quarkusTools.welcome",
     "workspaceContains:**/src/main/resources/application.properties",
     "workspaceContains:**/src/main/resources/application-?*.properties",
     "workspaceContains:**src/main/resources/META-INF/microprofile-config.properties",
@@ -43,12 +41,7 @@
     "workspaceContains:**/src/main/resources/application.yml",
     "workspaceContains:**/src/main/resources/templates",
     "onLanguage:microprofile-properties",
-    "onLanguage:quarkus-properties",
-    "onLanguage:java",
-    "onLanguage:qute-html",
-    "onLanguage:qute-json",
-    "onLanguage:qute-yaml",
-    "onLanguage:qute-txt"
+    "onLanguage:java"
   ],
   "main": "./dist/extension",
   "extensionDependencies": [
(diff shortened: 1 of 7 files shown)


src/test/vscodeTest/suite/index.ts
```diff
@@ -11,27 +11,14 @@ export function run(): Promise<void> {
...
-          } else {
-            c();
-          }
-        });
-      } catch (err) {
-        e(err);
-      }
-    });
+  // Run the mocha test
+  mocha.run(failures => {
```
datho7561: `mocha.run` will return before the callback is run, which means the program will exit without reporting any passes or failures. To get around this, you need to wrap it in a Promise and await that promise.

See https://github.com/redhat-developer/vscode-quarkus/pull/648/files#diff-5116d4bcabdc6045aa26cc1afdd56ab3155276c7d69ff9b3be07b9fb157da247R19

datho7561 (changes requested):

src/test/vscodeTest/suite/index.ts
fbricon: right. I still suck at Promises ;-)

package.json
```diff
@@ -34,21 +34,14 @@
...
     "workspaceContains:**/src/main/resources/application.yaml",
     "workspaceContains:**/src/main/resources/application.yml",
     "workspaceContains:**/src/main/resources/templates",
     "onLanguage:microprofile-properties",
-    "onLanguage:quarkus-properties",
-    "onLanguage:java",
-    "onLanguage:qute-html",
-    "onLanguage:qute-json",
-    "onLanguage:qute-yaml",
-    "onLanguage:qute-txt"
```
fbricon, line 51: onCommand and onLanguage events are automatically registered, no need to declare them explicitly here

src/test/vscodeTest/suite/extension.test.ts
```diff
@@ -10,17 +10,4 @@ describe('VS Code extension tests', () => {
...
-    const activationCommands = packageJSON.activationEvents.filter((s) => s.startsWith('onCommand:'))
-        .map((s) => s.substring('onCommand:'.length));
-    const QUARKUS_ACTIVATION_COMMANDS: string[] = [
-      // from package.json
-      VSCodeCommands.CREATE_PROJECT,
-      VSCodeCommands.QUARKUS_WELCOME,
-    ];
-
-    expect(activationCommands).to.have.members(QUARKUS_ACTIVATION_COMMANDS);
-  });
```
fbricon, line 25: onCommand events are automatically registered, no need to test them here

fbricon:
tests finally pass

package.json
```diff
@@ -442,14 +435,15 @@
     "@types/chai-fs": "^2.0.2",
     "@types/ejs": "^3.0.5",
     "@types/fs-extra": "^7.0.0",
+    "@types/glob": "^8.1.0",
```
datho7561: glob 9+ comes with the type declarations built into the main package, so we don't need this package.

datho7561 (changes requested):
I think this is ready to merge after this comment is addressed

package.json
fbricon: For some reason it didn't find the types until I added them

datho7561: It seems to work for me. What was reporting the error, VS Code or compiling from command line?

fbricon: VS Code. Will try again tomorrow

fbricon: it works now

datho7561 (approved):
Looks good. Thanks, Fred!

PR #672, opened 2024-02-08
Upgrade to @vscode/test-electron

datho7561:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
         "@types/vscode": "^1.82.0",
         "@types/which": "^3.0.4",
         "@types/yauzl": "^2.10.3",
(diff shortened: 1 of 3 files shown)


datho7561:
I think that something is wrong with the VS Code CDN (or maybe my internet connection) because downloading VS Code 1.86.1 as a part of the tests either takes a really long time for me, or the checksum doesn't match.

As well, I have VS Code 1.86.0 installed and no updates are yet available from the `yum` repo.

**UPDATE** not just on my end, checksum mismatch on GitHub Actions as well.

**UPDATE** passing locally for me now, but not yet working on CI it seems.

datho7561:
Going to just go ahead and merge this.

Issue #680, opened 2024-02-20
Missing extension command in VSCode


MintakaB:
Hi,
It was just to note that I can not see the command "Quarkus: Add extensions to current project:" in the drop down list.

OS: MacOS

VSCode:
Version: 1.86.2 (Universal)
Commit: 903b1e9d8990623e3d7da1df3d33db3e42d80eda
Date: 2024-02-13T19:42:13.651Z
Electron: 27.2.3
ElectronBuildId: 26908389
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Darwin x64 23.2.0

Quarkus Extension version: 1.17.0

![Screenshot 2024-02-20 at 18 38 39](https://github.com/redhat-developer/vscode-quarkus/assets/53517366/fdff36e2-0cff-400a-b291-44c73513e8a0)

fbricon:
Do you have a Quarkus project opened?

MintakaB:
hi, Yes, this is happening in all my three workspaces. It happens the same thing. No command to add new extensions to the project. It's a bit weird because it worked in the past!. Thank you for looking at the issue.

![Screenshot 2024-02-21 at 11 40 15](https://github.com/redhat-developer/vscode-quarkus/assets/53517366/cc017184-cca3-427d-af1a-b9ee4e6aa7cc)

mproep:
Hi, I think the http client from "axios", which is used in requestUtil.ts has some problem with https over http proxy. The problem is mentionend here: https://github.com/axios/axios/issues/4531

For me it also the reason why the vscode-quarkus extension is failing in the last step of "Generate a Quarkus project" with the same message mentioned here: https://github.com/redhat-developer/vscode-quarkus/issues/292

It was working for me eigth month ago.

Chealer:
Does this persist in VS Code 1.92+? If so, please provide a screenshot of a search for "quarkus:" in the keyboard shortcuts tab. For reference, with 1.17.0: ![image](https://github.com/user-attachments/assets/bcf570cf-7c6e-46f6-871b-db2f6e6253b4)

mproep:
For me the problem still persists.

angelozerr:
@datho7561 have you an idea why we have this issue?

Chealer:
<strike>

> For me nothing, the problem still persists
</strike>

<strike>I cannot make sense of what you wrote... thanks anyway, but a screenshot would be clearer.</strike>

[...]

PR #711, opened 2024-04-05
Replace axios library with built-in fetch

datho7561:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,6 @@
         "@types/yauzl": "^2.10.3",
         "@typescript-eslint/eslint-plugin": "^8.0.1",
         "@typescript-eslint/parser": "^8.0.1",
-        "axios": "^1.6.7",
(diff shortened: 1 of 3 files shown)


datho7561:
axios will still be pulled in as a transitive dependency from `@redhat-developer/vscode-redhat-telemetry`, but I've also opened a PR there to address this.

datho7561:
@angelozerr this might fix the issue with the proxy described in https://github.com/redhat-developer/vscode-quarkus/issues/680#issuecomment-2004067318 (This is not the bug that the person filing the issue encountered, but it should be addressed).

angelozerr:
@datho7561 I will test your PR without proxy. 

If it is working can I merge it?

datho7561:
Okay sounds good

angelozerr:
It works like a charm (tested without proxy), thanks @datho7561 !

PR #723, opened 2024-04-17
Migrate Jenkins release job to GitHub Actions.

rgrunber:
- Preserve existing Jenkinsfile as a fallback

This is nearly identical to https://github.com/redhat-developer/vscode-microprofile/blob/master/.github/workflows/release.yml with lsp4mp replaced with quarkus-ls & vscode-microprofile -> vscode-quarkus.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,158 @@
+name: release
+
+on:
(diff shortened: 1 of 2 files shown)


datho7561 (approved):
Seems good to me

PR #742, opened 2024-05-15
Use category for all commands and hide duplicated `Debug` command

### LINKED ISSUE #741: Use category `Quarkus` for all commands
currently, there is `Quarkus: ` as a prefix for most of the commands. In the Ui, it looks then similar to using a `category` but it is not leveraging the full power of VS Code API.

note: missing the `Quarkus:` for `Debug Quarkus` command

### LINKED ISSUE #743: Hide `Debug Quarkus` command from palette
it is used with a shorter name for editor quick menu. In command palette there is a more precise naming already `Debug current Quarkus project`

apupier:
Use category for all commands instead of using string prefix

fixes #741
fixes #743

### DIFF
--- a/package.json
+++ b/package.json
@@ -342,6 +342,10 @@
           "when": "quarkusProjectExistsOrLightWeight"
         },
         {
+          "command": "quarkusTools.debugQuarkusProject.short",
+          "when": "false"
+        },
+        {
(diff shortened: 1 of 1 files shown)


fbricon:
Thanks @apupier!

PR #815, opened 2024-08-21
fix: set platformOnly to false to collect all extensions in the wizard

angelozerr:
This PR allows to show Quarkus extensions without the platformOnly filter which allows to show langChain4j extensions:

image


fbricon:
Thanks @angelozerr!

PR #817, opened 2024-08-22
Force language to 'quarkus-properties' doesn't work for application.properties

angelozerr:
Force language to 'quarkus-properties' doesn't work for application.properties

To test this PR, install Spring extension which will set the application.properties to Spring Boot:

image

After loading the Java project, the application.properties language must be forced to Quarkus, but it doesn't work.

The PR fixes the regexp to accept only `application.properties`

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -116,7 +116,7 @@ function displayWelcomePageIfNeeded(context: ExtensionContext): void {
   }
 }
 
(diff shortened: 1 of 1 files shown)


datho7561 (approved):
Confirmed I run into this bug and this change fixes it. Thanks, Angelo!

PR #1122, opened 2025-09-16
Remove find-up dependency

datho7561:
It's a transitive dependency, so it shouldn't be listed as a direct dependency.


datho7561:
wait no i'm wrong

PR #1123, opened 2025-09-18
feat: Provide Qute snippet debugger

angelozerr:
feat: Provide Qute snippet debugger

Open a launch.json, you should have Qute debugging completion:

image

After completion apply:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -221,7 +221,19 @@
               }
             }
           }
-        }
+        },
+         "configurationSnippets": [
+          {
+            "label": "Qute: Debug Qute templates",
+            "description": "Debug Qute templates from Quarkus Application",
+            "body": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -221,7 +221,19 @@
               }
             }
           }
-        }
+        },
+         "configurationSnippets": [
+          {
+            "label": "Qute: Debugging Templates",
```
datho7561: I think "Qute: Debug Qute templates" might be more clear for the label and name, let me know what you think

datho7561 (changes requested):

package.json
angelozerr: fixed

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #1125, opened 2025-09-18
Activate vscode-quarkus when Qute debug is launched.

angelozerr:
This PR activate vscode-quarkus when user click on a launch which defines qute type debug adapter without waiting start of JDT LS or having src/main/resources/templates directory.


datho7561 (approved):

angelozerr:
Thanks @datho7561 !

PR #1194, opened 2026-01-06
Define qute.inlayHint.showMessages

angelozerr:
Define qute.inlayHint.showMessages

### DIFF
--- a/package.json
+++ b/package.json
@@ -338,6 +338,11 @@
           "default": true,
           "markdownDescription": "Show section parameter default value. Default is `true`."
         },
+        "qute.inlayHint.showMessages": {
+          "type": "boolean",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -338,6 +338,11 @@
           "default": true,
           "markdownDescription": "Show section parameter default value. Default is `true`."
         },
+        "qute.inlayHint.showMessages": {
```
datho7561, line 341: Also add it here: https://github.com/redhat-developer/vscode-quarkus/blob/master/docs/qute/README.md#settings

angelozerr: fixed

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #1195, opened 2026-01-07
Use npx

angelozerr:
Use npx


datho7561 (approved):
Seems like a very good idea, it should do the same thing as before on Linux and macOS, and it will work on Windows.

PR #1197, opened 2026-01-12
(Re)Bind LSP request 'qute/template/projects'

angelozerr:
(Re)Bind LSP request 'qute/template/projects'

I have noticed that the LSP request  'qute/template/projects' binding has been removed from 2 years!

We need this LSP request to load Qute projects when Qute LS starts (you should see Loading in the progress 

image


datho7561 (approved):
Thanks, Angelo! I see the request and response in the logs.

PR #1289, opened 2026-08-20
Update vulnerable transitive dependencies

datho7561:


datho7561:
Seems to work locally for me, and tests pass

